PR #1, opened 2022-09-27
Attempting to unify batched and unbatched tensors

hugary1995:
Unify batched/unbatched Scalar
Unify batched/unbatched SymR2

Several renamings

Some additional unit tests


hugary1995:
I'm happy with the +130 -198 line changes :-)

I haven't attempted to unify the SymSymR4 yet, but that's definitely doable...with more templating.

@reverendbedford This could be something fun for you to look at during the conference.

reverendbedford:
I’ll take a look, right now I’m working on setting up all the structure for
our standard models.

hugary1995:
The default (empty) constructors are giving me headaches: For unbatched tensors, we could initialize zeros. But for batched tensors, we can't do anything as the batch size is dynamic.

I could either (1) throw runtime errors for the batched empty constructor, or (2) make the batch sizes (not only the batch dimensions) static.

hugary1995:
Alternatively we could delete the empty constructors. Users just won't be able to write
```cpp
Scalar E, nu;
SymR2 stress, strain;
```
etc

which probably isn't a big deal (?)

hugary1995:
Okay, I've thought about it more carefully. I think we need to first reach a consensus on the syntax we intend to support, then everything could be done at the FixedDimTensor level (including dot products and tensor operations). The Scalar, SymR2 and SymSymR4 can be typedefs.

For mixing tensor products, e.g. batched dot batched, batched dot unhatched, etc., we should be able to figure out whether each of the operand is batched at compile time, then there will only be two cases:
- unbatched dot unbatched: use the `matmul` routine.
- all other scenarios: we first manually broadcast the unbatched operands, then use the `bmm` routine.

reverendbedford:
Keep in mind we will need to support more than just the "standard" dot product in the end.  Outer products of the type SymR2 x SymR2 -> SymSymR4 I'll need even for small strain models and then things only get weirder for crystal plasticity.  We will also have batched quaternions for crystal plasticity where there isn't a standard dot product.  Does that change any of your plans? 

I'm plugging away at getting a "real" material model (J2 viscoplasticity with some simple hardening) setup.

hugary1995:
I need to dig into the torch apis to figure that out. The best scenario would be that torch has predefined batched versions of the more complicated operations. If not, we'll have to write cuda programs on our own, I guess. So that probably won't change my plan.

hugary1995:
A bunch of reorganizations: now the data structures are
- BatchTensor: static batch dimension, dynamic batch size, dynamic base size
- FixedDimTensor: static batch dimension, dynamic batch size, static base size
- Scalar/SymR2/SymSymR4: static batch dimension of 1, dynamic batch size, static base size

Note that I am no longer distinguishing between "batched" and "unbatched" -- tensors will always be batched, but it can be "unbatched" in the sense that the batch size is 1. This choice has several advantages:
1. The intermediate layer `Standard(Un)BatchedTensor`s can be removed.
2. Now batched and unbatched versions of the tensors are unified without templating nor `enable_if`s, which are getting annoying as more operations are added.
3. Broadcasting batch operations using torch becomes trivial (with the help of `torch::einsum`). Take a look at `SymSymR4::operator*`. For operations that cannot be expressed using einsum, we'll have to write our own CUDA kernels anyways.

I also added new vector-based constructors over all batches, so initialization should be much easier. Something like
```cpp
SymR2 a({1, 2, 3, 4, 5, 6}); // This will initialize a SymR2 with the specified filled values.
SymR2 b({1, 2, 3, 4, 5, 6}, 1000); // Same as above, but the same values are filled over all 1000 batches.
```
similarly
```cpp
Scalar a(5.5); // This is a single scalar. The underlying torch::Tensor has shape (1, 1).
Scalar b(5,5, 1000); // This is a batched scalar. The underlying torch::Tensor has shape (1000, 1).
SymSymR4 c({1, 2, 3, 4, 5, ..., 36}, 1000); // This is the plain initializer for a SymSymR4. In the future we could handle this initialization in a smarter way.
```

I'll try to unify the other branch of features, the labeled tensors, in another PR. For now at least their unit tests still pass...

reverendbedford:
I’m going to keep going on the “feature” branch until plasticity works.
Then one of us will have to circle back and merge…. I’ll try not to add too
much math.

hugary1995:
lol, now we are at +234 -513 line change. I'm fairly happy with the `FixedDimTensor` side of things now. I guess I'll just wait for your plasticity update and go from there.

hugary1995:
Sorry this is getting out of the scope of this PR, but... the last commit demonstrates the usage of backward AD to get dstate/dforce. Speaking of speed, I think this is the fastest libtorch can get us.

hugary1995:
I'm gradually adding methods and operators for the tensors. Feel free to give me a list of methods (for your plasticity update).

reverendbedford:
Thus far I've used:

SymSymR4:
- identity and deviatoric identity static named constructors
- dot -> "ijkl,klmn" with another SymSymR4

SymR2:
- zeros and identity static named constructors
- "tr" and "dev"
- "double_contraction" -> "ij,ij" with another SymR2
- "norm" -> sqrt(double_contraction(*this))
- "outer" -> "ij,kl" with another SymR2

Scalar:
- named constructor "constant" which repeats a constant value across the batch dimension

Plus all addition/subtraction and scalar multiplication operations.  These all "magically" work right now by providing the converting constructors from `torch::Tensor`, let's try to keep it that way...

I'm refactoring most of the actual model stuff that I already wrote to inherit from a class providing an interface for a function that takes a variable number of `State` objects and input and maps those to a single `State` output.  The class will provide a default derivative implementation that uses AD to return the derivative of the function with respect to each state object.  But then you can override that with a hard coded derivative.

The only problem right now are stupid yield surfaces as always because you need second derivatives.

hugary1995:
The `These` in

>These all "magically" work right now by providing the converting constructors from torch::Tensor, let's try to keep it that way...

means

>all addition/subtraction and scalar multiplication operations

right? I don't see how things like `tr` and `dev` would magically work on batched tensors.

reverendbedford:
Yes that only applies to scalar multiplication and simple addition and
subtraction.

hugary1995:
Note to myself: right now 18 out of the 24 tests fail if I configure to use GPU. The fix is probably trivial, but this is something I'll have to look at after I'm done with the operators.

hugary1995:
:smiley: :smiley: :smiley: :smiley: :smiley: :smiley: 
image

hugary1995:
All methods you listed above are there. They are all tested with full power, e.g. the full combination of (unbatched,unbatched) (unbatched,batched) (batched,unbatched) (batched,batched).

reverendbedford:
Cool, still trying to work on the plasticity models during breaks.  Slowly getting there...

hugary1995:
Is `actual_models` the branch you are working on? I can start (manually) merging some of your stuff into my branch so that there is less risk of merging two completely diverged branches in the end.

reverendbedford:
Yes

Issue #16, opened 2022-11-01
Add license, appropriate source file copyright statements


reverendbedford:
Ugh, we will need to do this as well before we move things over.  I'll file a software disclosure now as it will take awhile for ANL to approve things.

hugary1995:
I'll write a python script to do handle source file copyright headings.

The script will scan all files with specified extension names, check if there are missing/incorrect headings by comparing to a provided template.
1. The script will exit with nonzero if any change is needed.
2. Offer a command line option to let the script make the correction.

The script should also be part of the github workflow for PRs.

hugary1995:
Once software disclosure is approved, we simply need to add the appropriate header to #40 and run the script.

PR #40, opened 2022-12-09
Add a script to check copyright heading

hugary1995:
for each source file, that is.

Also added a github workflow to enforce it. The workflow should fail now. But when we are ready, I'll run the script locally and add all the appropriate headings.

ref #16


hugary1995:
closed in favor of #61

PR #123, opened 2024-02-02
Add GTN model to solid_mechanics

reverendbedford:
Adds the basic version of the Gurson-Needleman-Tvergaard model, which consists of a non-J2 yield surface plus an evolution equation for the void fraction.  For the moment I didn't implement the traditional void nucleation or coalescence terms, as we're going to be using this "in reverse" to model compaction for the time being.

Verified the results against WARP3D as a reference implementation.


hugary1995 (approved):
I'll merge once the tests pass.

PR #168, opened 2024-08-23
Got jit to work

hugary1995:


hugary1995:
closing now as there's apparently much more to do...

PR #173, opened 2024-09-24
Test scalar trial stress update

dschwen:
Run this through testing....


dschwen:
The new test now runs with
```
  Result has extra variable input.state/se.
  Result has extra variable output.forces/s.
  Result has wrong value for variable output.state/ep_rate. Maximum mixed
  difference = 1.255339e-01 > atol = 1.000000e-08
  Result is missing variable output.state/s.
```

I had to change `state/s` to `forces/s` to make it work, but that that breaks the expected outputs.  I wonder why the `state/ep_rate` result is so wrong?

hugary1995:
As long as `state/S` (the stress tensor) matches the other version, I'd call this good.

include/neml2/models/ScalarTrialStressUpdate.h
reverendbedford, line 1: Name should probably indicate the restrictions of when this can be used (isotropic elasticity, J2 flow function).  I also mentioned to Gary that I'd like some way to enforce those restrictions, i.e. throw an error if you're not using isotropic elasticity and the J2 yield surface.  Maybe he has some idea how to do that.

src/neml2/models/ScalarTrialStressUpdate.cxx
```diff
@@ -0,0 +1,55 @@
...
+    _inelastic_strain(declare_input_variable<Scalar>("inelastic_strain")),
+    _inelastic_strain_old(declare_input_variable<Scalar>(_inelastic_strain.name().old())),
+    _updated_trial_stress(declare_output_variable<Scalar>("updated_trial_stress")),
+    _E(declare_parameter<Scalar>("E", "youngs_modulus", /*allow_nonlinear=*/true)),
+    _nu(declare_parameter<Scalar>("nu", "poisson_ratio", /*allow_nonlinear=*/true))
+{
+}
+
+void
+ScalarTrialStressUpdate::set_value(bool out, bool /*dout_din*/, bool /*d2out_din2*/)
```
reverendbedford: Don't we need the chain rule derivatives here?

reverendbedford (changes requested):

src/neml2/models/ScalarTrialStressUpdate.cxx
dschwen: Yeah, we do. I though I had added use ad options here...

dschwen: And I did. What are you saying is missing then?

reverendbedford: Before we merge it would be nice to have the analytic derivatives included, like all the other models.

dschwen: I implemented the derivatives

dschwen:
The code coverage test seems broken ("Bad credentials")

reverendbedford:
Alright, let's figure out why the stupid action won't run.

hugary1995:
We ran into this issue with Lynn's PR before. I think we ended up making him a collaborator so the tests could run. But maybe we should figure out how to get the tests to run for external PRs.

hugary1995:
@dschwen we shall add unit tests for the new models. I'll take care of those.

reverendbedford:
> We ran into this issue with Lynn's PR before. I think we ended up making him a collaborator so the tests could run. But maybe we should figure out how to get the tests to run for external PRs.

This is indeed the problem, I think it fails when you're submitting as a PR from a fork, rather than directly on this repo.  I added @dschwen as a collaborator on this project but this is still setup as a PR from his fork.

Let's see if I can fix it...

hugary1995:
We need to set `deploy-repository` and `token` for the documentation preview to be deployed. See https://github.com/marketplace/actions/deploy-pr-preview

reverendbedford:
@dschwen could you please move this PR over to our new location here: https://github.com/applied-material-modeling/neml2  @hugary1995 claims the tests will also run fine now from your fork.

hugary1995:
> @dschwen could you please move this PR over to our new location here: https://github.com/applied-material-modeling/neml2 @hugary1995 claims the tests will also run fine now from your fork.

Not yet... hold on to this PR

reverendbedford:
@dschwen in case you couldn't tell we just moved the whole repository.  So it just moved it automatically.

hugary1995 (approved):

include/neml2/models/ScalarTrialStressUpdate.h
hugary1995: Name is descriptive enough now.

> Maybe he has some idea how to do that.

I think "action" will take care of that, eventually...

reverendbedford (approved):

include/neml2/models/ScalarTrialStressUpdate.h
reverendbedford: Ya, okay the name is enough for now.

PR #175, opened 2024-12-03
Merge applied-material-modeling

hugary1995:
Bump version to 2.0.0

Also updates workflows to grant correct permissions to GITHUB_TOKEN. PRs from forks should work as expected now.

All external PRs now require approval from the neml2-dev team before the checks can run. This is because some workflows require elevated priviledges and is subject to malicious exfiltration.

Allow posting multiple comments in PR The sticky PR comment action allows the use of different headers to distinguish comments from each other. This is used to post multiple comments in a PR.

Remove support for pdf documentation
We no longer need the pdf manual as the software has been released. Maintaining the latex templates, headers, and footers is an overhead.

Do not provide scalartype caster if libtorch has it For torch version > 2.2.4, libtorch provides the pybind11 `type_caster` for `torch::ScalarType`. For backward compatibility, we will provide our version of the type_caster for older versions of torch.

Update repository link, pytorch version, and python version

Update macos runner

Update readme badges


reverendbedford (approved):

reverendbedford:
Will talk with you about the tests before I merge.

PR #177, opened 2024-12-04
Use out-of-place evaluation wherever possible

### LINKED ISSUE #165: More option metadata
Two that are needed sooner are
1. If an option is properly "defined" -- this can be set when the `operator=` is used on an option.
2. If an option has been modified by the user -- this metadata will be set by the parser.

### LINKED ISSUE #180: freeze lookback for NEML2-Pyzag interface
`neml2.pyzag.NEML2PyzagModel.lookback` should be fixed to 1.

hugary1995:
- On-demand matrix assembly
- Fix unit tests, regression tests, and verification tests
- Add more model forward operator APIs
- Refactor LabeledAxis
- Update unit test and drivers to use the new model evaluation APIs


hugary1995:
This will also close #165 and #180

hugary1995:
I'm pretty sure this is my personal record in # line change.

doc/content/system/model.md
```diff
@@ -109,6 +109,82 @@ The assembly of the partial derivatives into the total derivative \f$\partial y
...
+```
+
+\note
+Each model can use a mix of hand-coded derivatives and AD derivatives. However, an error will be raised if hand-coded derivatives are provided for those marked by neml2::Variable::request_AD.
+
+Since a composed model uses chain rule to efficiently evaluate the total derivatives, automatic differentiation is disabled for `ComposedModel`. However, each of the child model can still use AD to calculate the _partial_ derivatives of its own forward operator. Moreover, AD and non-AD models can be composed together.
+
+## Model assembly {#model-assembly}
+
+NEML2 stores each variable in contiguous memory, but does not guarantee contiguity across variables. This choice is made to allow for massive asynchronous evaluation (with the help of lazy tensor) and to reduce memory consumption (since variables can have different number of batch shapes). However, this choice is not ideal for a family of nonlinear material models whose constitutive updates require solving one (or more) implicit system of equations. To address such issue, NEML2 offers two mechanisms to facilitate the creation of the implicit system (e.g., its residual and Jacobian):
```
reverendbedford:
```suggestion
NEML2 stores each variable in contiguous memory, but does not guarantee contiguity across variables. This choice is made to allow for massive asynchronous evaluation (with the help of lazy tensors) and to reduce memory consumption (since variables can have different number of batch shapes). However, this choice is not ideal for a family of nonlinear material models whose constitutive updates require solving one (or more) implicit system of equations. To address such issue, NEML2 offers two mechanisms to facilitate the creation of the implicit system (e.g., its residual and Jacobian):
```

doc/content/system/model.md
```diff
@@ -109,6 +109,82 @@ The assembly of the partial derivatives into the total derivative \f$\partial y
 
 ## Automatic differentiation {#automatic-differentiation}
 
-Deriving and implementing derivatives of the forward operator can be cubersome from times to times. NEML2 offers the option to use automatic differentiation to obtain derivatives. To enable automatic differentiation, simply set the `_use_AD_first_derivative` and/or the `_use_AD_second_derivative` options to `true`.
+Deriving and implementing derivatives of the forward operator can be cubersome from times to times. NEML2 offers the option to use automatic differentiation (AD) to obtain derivatives. To enable automatic differentiation, one simply need to override the neml2::Model::request_AD method and specify which derivatives should be computed using AD:
```
reverendbedford:
```suggestion
Deriving and implementing derivatives of the forward operator can be cumbersome from times to times. NEML2 offers the option to use automatic differentiation (AD) to obtain derivatives. To enable automatic differentiation, one simply need to override the neml2::Model::request_AD method and specify which derivatives should be computed using AD:
```

doc/content/system/model.md
```diff
@@ -109,6 +109,82 @@ The assembly of the partial derivatives into the total derivative \f$\partial y
...
+### Axis labeling {#axis-labeling}
+
+NEML2 provides a data structure named [LabeledAxis](@ref neml2::LabeledAxis) to create a contiguous layout for scattered input/output variables. Typically, each model contains an input axis for input variables and an output axis for output variables.
+
+The [LabeledAxis](@ref neml2::LabeledAxis) contains all information regarding how the variables of interest should be contiguously laid out. In other words, the labeled axis maintains the mapping between variables and their contiguous slice along an axis. The following naming convention is used:
+- Item: A labelable slice of data
+- Variable: An item that is also of a [NEML2 primitive tensor type](@ref tensor-types)
+- Sub-axis: An item of type `LabeledAxis`
+
+So yes, an axis can be labeled recursively, e.g.,
```
reverendbedford:
```suggestion
An axis can be labeled recursively, e.g.,
```

include/neml2/contrib/boost/bitmask_operators.h
reverendbedford, line 1: Already asked, but do we need this now?

include/neml2/misc/math.h
```diff
@@ -40,11 +40,20 @@ constexpr Real eps = std::numeric_limits<at::scalar_value_type<Real>::type>::eps
 constexpr Real sqrt2 = 1.4142135623730951;
 constexpr Real invsqrt2 = 0.7071067811865475;
 
-constexpr Size mandel_reverse_index[3][3] = {{0, 5, 4}, {5, 1, 3}, {4, 3, 2}};
-constexpr Size mandel_index[6][2] = {{0, 0}, {1, 1}, {2, 2}, {1, 2}, {0, 2}, {0, 1}};
-
-constexpr Size skew_reverse_index[3][3] = {{0, 2, 1}, {2, 0, 0}, {1, 0, 0}};
-constexpr Real skew_factor[3][3] = {{0.0, -1.0, 1.0}, {1.0, 0.0, -1.0}, {-1.0, 1.0, 0.0}};
+constexpr std::array<std::array<Size, 3>, 3> mandel_reverse_index{
```
reverendbedford, line 43: Doesn't really matter, but why this change?

include/neml2/misc/types.h
```diff
@@ -42,6 +43,88 @@ using TensorIndices = torch::SmallVector<TensorIndex>;
...
+  torch::Tensor as_tensor() const;
+};
+
+/// Comparison operators
+///@{
+bool operator==(const TraceableTensorShape & lhs, const TraceableTensorShape & rhs);
+bool operator!=(const TraceableTensorShape & lhs, const TraceableTensorShape & rhs);
+///@}
+
+/**
```
reverendbedford, line 109: I start to worry we have a lot of magic numbers/strings going on with the canonical axis labels.  I think you do some magic strings in `Variable` as well, though I guess I'll get to that when I find it.  Is there a way to collect all these definitions in one file, in case we want to make global changes to their names later?

include/neml2/solvers/NonlinearSystem.h
```diff
@@ -36,10 +36,80 @@ namespace neml2
 class NonlinearSystem
 {
 public:
-  static OptionSet expected_options();
+  /**
+   * Convenience struct to hold residual to prevent developers from accidentally confusing the
+   * scaled and unscaled residual
+   */
+  template <bool scaled>
+  struct RES : public Tensor
```
reverendbedford: Real picky, but do we want these to be all caps or do we want `Res`, `Jac`, etc.?

include/neml2/tensors/TensorBaseImpl.h
reverendbedford, line 1: There are like no comments in this header...

python/tests/pyzag/test_NEML2PyzagModel.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # Modify again, batch shape = ()
+    pmodel.elasticity_E.data = torch.tensor(1.3e5)
+    pmodel._update_parameter_values()
+    assert torch.allclose(pmodel.model.elasticity_E.torch(), pmodel.elasticity_E)
+    assert pmodel.model.elasticity_E.tensor().batch.shape == ()
+
+
+@pytest.mark.parametrize("input", ["elastic_model", "viscoplastic_model", "km_mixed_model"])
+# @pytest.mark.parametrize("input", ["elastic_model", "viscoplastic_model"])
```
reverendbedford: Don't need the comment.

include/neml2/models/solid_mechanics/crystal_plasticity/CrystalMean.h
```diff
@@ -22,30 +22,35 @@
...
+  CrystalMean(const OptionSet & options);
 
-  return true;
-}
-}
+protected:
+  void set_value(bool out, bool dout_din, bool d2out_din2) override;
+
+  /// Crystal geometry class with slip geometry
+  const crystallography::CrystalGeometry & _crystal_geometry;
```
reverendbedford, line 48: Why do we need this?

include/neml2/models/Assembler.h
```diff
@@ -24,28 +24,63 @@
...
+  VectorAssembler(const LabeledAxis & axis)
+    : _axis(axis)
+  {
+  }
+
+  /// Assemble a vector of vectors (by variables)
+  Tensor assemble(const ValueMap &) const;
+
+  /// Split the vector (by variables)
+  ValueMap disassemble(const Tensor &) const;
```
reverendbedford: Are these the best names?  Maybe something like `split_by_subaxis` and `split_by_variable` (or `assemble_by_...`).

include/neml2/models/Variable.h
```diff
@@ -0,0 +1,360 @@
...
+#include "neml2/models/map_types.h"
+
+namespace neml2
+{
+// Forward declarations
+class Model;
+class Derivative;
+enum class TensorType : int8_t;
+
+class VariableBase
```
reverendbedford, line 46: Might be in the specialized classes, but a top-level comment would b nice.

include/neml2/models/Variable.h
```diff
@@ -0,0 +1,360 @@
...
+           Model * owner,
+           TensorShapeRef list_shape,
+           TensorShapeRef base_shape)
+    : VariableBase(std::move(name_in), owner, list_shape),
+      _base_sizes(base_shape),
+      _ref(nullptr),
+      _ref_is_mutable(false)
+  {
+  }
+
```
reverendbedford, line 252: Some of these are obvious by the name, but many aren't.  More comments will help.

include/neml2/models/Variable.h
reverendbedford, line 1: `VariableBase` is a giant class, I don't know if there is a way to split things up a bit.  Maybe there isn't, if so I won't die on this hill.

src/neml2/models/Variable.cxx
```diff
@@ -0,0 +1,425 @@
...
+}
+
+Model &
+VariableBase::owner()
+{
+  neml_assert_dbg(_owner, "Owner of variable '", name(), "' has not been defined.");
+  return *_owner;
+}
+
+bool
```
reverendbedford, line 52: This is another example of magic strings defining our canonicle axes...  Can we put these types of definitions in one file to make it easy to change things later?

src/neml2/models/LabeledAxis.cxx
```diff
@@ -0,0 +1,494 @@
...
+
+    if (*other._subaxes.at(name) != *axis)
+      return false;
+  }
+
+  return true;
+}
+
+void
+LabeledAxis::cache_reserved_subaxis(const std::string & axis_name)
```
reverendbedford, line 431: Same comment on centralizing definitions of the canonical axes names.

src/neml2/models/LabeledAxisAccessor.cxx
```diff
@@ -69,38 +75,52 @@ LabeledAxisAccessor::prepend(const LabeledAxisAccessor & axis) const
...
 LabeledAxisAccessor::start_with(const LabeledAxisAccessor & axis) const
 {
-  return slice(0, axis.size()) == axis;
+  return slice(0, int64_t(axis.size())) == axis;
+}
+
+LabeledAxisAccessor
+LabeledAxisAccessor::current() const
+{
+  neml_assert(_item_names.size() >= 1, "variable name length must be at least 1");
```
reverendbedford, line 154: More magic strings to define the canonical axes.

src/neml2/models/LabeledAxisAccessor.cxx
```diff
@@ -69,38 +75,52 @@ LabeledAxisAccessor::prepend(const LabeledAxisAccessor & axis) const
...
+    return remount("forces");
+  throw NEMLException("Unable to find current counterpart of variable named '" +
+                      utils::stringify(*this) + "'");
 }
 
 LabeledAxisAccessor
 LabeledAxisAccessor::old() const
 {
-  // NOLINTNEXTLINE(readability-container-size-empty)
   neml_assert(_item_names.size() >= 1, "variable name length must be at least 1");
```
reverendbedford, line 166: And here.

python/neml2/pyzag/interface.py
```diff
@@ -65,291 +37,221 @@ class NEML2PyzagModel(nonlinear.NonlinearRecursiveFunction):
...
+    @lookback.setter
+    def lookback(self, lookback):
+        if lookback != 1:
+            raise ValueError("NEML2 models only support lookback of 1")
+        self._lookback = lookback
 
-    def _setup_parameters(self, exclude_parameters):
-        """Mirror parameters of the neml model with torch.nn.Parameter
+    @property
+    def nstate(self):
```
reverendbedford, line 72: Assume you find a way to centralize the string names of the canonical axes, we can wrap that in python to provide these values, rather than the hard-coded strings.

python/neml2/pyzag/interface.py
```diff
@@ -65,291 +37,221 @@ class NEML2PyzagModel(nonlinear.NonlinearRecursiveFunction):
...
+        """Simple consistency checks, could be a debug check but we only call this once"""
 
-        # We may need to update the batch shapes, so this lambda sorts out the correct shape
-        def make_tensor(orig_name, neml_name):
-            new_value = getattr(self, orig_name)
-            current_value = self.model.get_parameter(neml_name)
-            batch_dim = new_value.dim() - current_value.tensor().base.dim()
-            return Tensor(new_value.clone(), batch_dim)
+        # First run diagnostics from NEML2
+        # neml2.diagnose(self.model)
```
reverendbedford, line 88: Why commented out?

python/neml2/pyzag/interface.py
```diff
@@ -65,291 +37,221 @@ class NEML2PyzagModel(nonlinear.NonlinearRecursiveFunction):
...
+                raise ValueError(errmsg.format(pname))
+            if pname in exclude_parameters:
+                continue
+            param.requires_grad_(True)
+            self.register_parameter(pname, torch.nn.Parameter(param.torch()))
 
-    def collect_state(self, tensor_dict):
-        """Assemble the state from a dictionary of tensors
+    def _update_parameter_values(self):
+        """Copy over new parameter values"""
```
reverendbedford, line 160: Suprised the linter doesn't complain.

reverendbedford (changes requested):
Several minor comments.  The biggest request is to centralize the names of the "canonical" axes to make those easier to configure later on.  Hopefully the detailed comments on this idea make it clear what I mean.

include/neml2/misc/math.h
hugary1995: Because clang-tidy suggests using std::array to replace c-arrays. I don't want to disable that check, so I switched to std::array here.

include/neml2/tensors/TensorBaseImpl.h
hugary1995: Cuz they are "implementation". The comments are in the header file "TensorBase.h". I'll add a comment pointing people to the actual header.

include/neml2/models/solid_mechanics/crystal_plasticity/CrystalMean.h
hugary1995: We need the crystal geometry object to know the number of slip systems. Alternatively, we can ask the user to specify nslip, but that's one additional input option.

Once "list" tensors are better handled, we could infer list shape, and then we can remove the need of crystal geometry in this class.

include/neml2/contrib/boost/bitmask_operators.h
hugary1995: Removed.

include/neml2/models/Variable.h
hugary1995: I'd argue that they are all overriding the base methods. By doxygen convention, the comments inherit from the base methods.

python/neml2/pyzag/interface.py
hugary1995: well this is allowed!

python/neml2/pyzag/interface.py
hugary1995: I added a comment explaining why.

include/neml2/models/Variable.h
hugary1995: It's a little counter-intuitive to split VariableBase. It serves as a polymorphic base so that we could store derived Variable<T> in a homogeneous container. So from that perspective, it's actually better to remain as a single class.

reverendbedford (approved):

PR #178, opened 2024-12-10
Integrate pyzag bindings for neml2

reverendbedford:
Add pyzag wrappers for NEML2 model and include some basic tests.

This will make pyzag a dependency of NEML2.


src/neml2/models/LinearInterpolation.cxx
```diff
@@ -42,35 +42,29 @@ LinearInterpolation<T>::expected_options()
...
 
 template <typename T>
 void
 LinearInterpolation<T>::set_value(bool out, bool dout_din, bool d2out_din2)
 {
+  const auto X0 = this->_X.batch_index({indexing::Ellipsis, indexing::Slice(indexing::None, -1)});
+  const auto X1 = this->_X.batch_index({indexing::Ellipsis, indexing::Slice(1, indexing::None)});
+  const auto Y0 = this->_Y.batch_index({indexing::Ellipsis, indexing::Slice(indexing::None, -1)});
+  const auto slope = math::diff(this->_Y, 1, this->_Y.batch_dim() - 1) /
+                     math::diff(this->_X, 1, this->_X.batch_dim() - 1);
```
hugary1995, line 57: This is suboptimal. Here's the issue for tracking that.

src/neml2/models/Model.cxx
```diff
@@ -452,10 +452,14 @@ void
...
   check_inplace_dbg();
+
+  if (is_AD_disabled())
   {
-    c10::InferenceMode guard(is_AD_disabled());
+    c10::InferenceMode guard;
     set_value(true, false, false);
   }
+  else
+    set_value(true, false, false);
```
hugary1995: Don't need these changes.

src/neml2/models/Model.cxx
```diff
@@ -469,8 +473,13 @@ Model::value_and_dvalue()
...
   {
-    c10::InferenceMode guard(is_AD_disabled());
-    set_value(true, true, false);
+    if (is_AD_disabled())
+    {
+      c10::InferenceMode guard;
+      set_value(true, true, false);
+    }
+    else
+      set_value(true, true, false);
```
hugary1995: Don't need these changes.

src/neml2/models/Model.cxx
```diff
@@ -491,8 +500,13 @@ Model::dvalue()
...
   {
-    c10::InferenceMode guard(is_AD_disabled());
-    set_value(false, true, false);
+    if (is_AD_disabled())
+    {
+      c10::InferenceMode guard;
+      set_value(false, true, false);
+    }
+    else
+      set_value(false, true, false);
```
hugary1995: Don't need these changes.

src/neml2/models/Model.cxx
```diff
@@ -514,8 +528,13 @@ Model::value_and_dvalue_and_d2value()
...
   {
-    c10::InferenceMode guard(is_AD_disabled());
-    set_value(true, true, true);
+    if (is_AD_disabled())
+    {
+      c10::InferenceMode guard;
+      set_value(true, true, true);
+    }
+    else
+      set_value(true, true, true);
```
hugary1995: Don't need these changes.

src/neml2/models/Model.cxx
```diff
@@ -546,8 +565,13 @@ Model::dvalue_and_d2value()
...
   {
-    c10::InferenceMode guard(is_AD_disabled());
-    set_value(false, true, true);
+    if (is_AD_disabled())
+    {
+      c10::InferenceMode guard;
+      set_value(false, true, true);
+    }
+    else
+      set_value(false, true, true);
```
hugary1995: Don't need these changes.

src/neml2/models/Model.cxx
```diff
@@ -577,8 +601,13 @@ Model::d2value()
...
   {
-    c10::InferenceMode guard(is_AD_disabled());
-    set_value(false, false, true);
+    if (is_AD_disabled())
+    {
+      c10::InferenceMode guard;
+      set_value(false, false, true);
+    }
+    else
+      set_value(false, false, true);
```
hugary1995: Don't need these changes.

src/neml2/models/SmoothLinearInterpolation.cxx
```diff
@@ -0,0 +1,109 @@
...
+  : Interpolation<T>(options),
+    _k(options.get<Real>("sharpness")),
+    _X0(this->template declare_buffer<Scalar>(
+        "X0", this->_X.batch_index({indexing::Ellipsis, indexing::Slice(indexing::None, -1)}))),
+    _X1(this->template declare_buffer<Scalar>(
+        "X1", this->_X.batch_index({indexing::Ellipsis, indexing::Slice(1, indexing::None)}))),
+    _Y0(this->template declare_buffer<T>(
+        "Y0", this->_Y.batch_index({indexing::Ellipsis, indexing::Slice(indexing::None, -1)}))),
+    _Y1(this->template declare_buffer<T>(
+        "Y1", this->_Y.batch_index({indexing::Ellipsis, indexing::Slice(1, indexing::None)})))
```
hugary1995: Shouldn't these be in the set_value method? i.e., when the parameter values are updated during training, they should be recalculated.

hugary1995 (changes requested):
Changes on the c++ side look okay to me. A few minor comments.

Now onto the python part...

src/neml2/models/LinearInterpolation.cxx
hugary1995: https://github.com/applied-material-modeling/neml2/issues/179

python/neml2/pyzag/interface.py
```diff
@@ -0,0 +1,333 @@
...
+        python_sep_char="_",
+        *args,
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+
+        self.model = model
+        self.lookback = (
+            1  # Hard coded because there aren't any other options in NEML2 right now
+        )
```
hugary1995, line 94: I guess this doesn't prevent people from modifying the attribute. Since python doesn't have a direct way of "freezing" an attribute, I think it's best to assert lookback == 1 at critical locations.

src/neml2/models/LinearInterpolation.cxx
reverendbedford: Okay doke

python/neml2/pyzag/interface.py
hugary1995, line 1: I will have to make a lot of changes for the interface after my PR anyways. So I just glanced through this file.

hugary1995 (changes requested):
That's all my comments.

python/neml2/pyzag/interface.py
hugary1995: Saving this for later: https://github.com/applied-material-modeling/neml2/issues/180

hugary1995 (approved):
We are running out of independent reviewers, so I'll just merge when all tests pass.

hugary1995:
Urgh the clang-tidy check is unbearably slow. We need to fix it...

reverendbedford:
Don't merge yet, I haven't rerun the stochastic example to make sure I got all the changing things.

requirements.txt
```diff
@@ -2,6 +2,8 @@
 ##### pip requirements ######
 #
 numpy<2
+pyzag
+nbdime
 -r doc/requirements.txt
 -r scripts/requirements.txt
 -r python/tests/requirements.txt
```
hugary1995:
```suggestion
pyzag
-r doc/requirements.txt
-r python/tests/requirements.txt
```

Then create `python/examples/requirements.txt` with
```
nbdime
pyro
tqdm
matplotlib
```

hugary1995 (changes requested):
Let's not add nbdime, pyro, and tqdm as a global dependency, as they bring in a fair number of packages.

python/examples/README.md
hugary1995, line 1: I know it's sorta obvious, but let's also mention that at least one notebook kernel should be installed: https://jupyter.org/install

hugary1995 (changes requested):

python/examples/README.md
hugary1995: Alternatively, `conda install ipykernel`.

hugary1995: Urgh, this is annoying. We also need ipywidgets to get rid of the tqdm warning. If the user installed jupyter with pip, then it's just
```
pip install ipywidgets
```
I think this should also be mentioned in the README.md if we want to avoid notebook diffs.

hugary1995 (changes requested):
My final comment:

I think torch.manual_seed doesn't guarantee reproducibility in different environments. The SVI example has two stochastic aspects: training and inference. If you save the checkpoint and reload it in the notebook, the trained model should be reproducible. But then IDK how to make the inference part (i.e., sampling the posterior) truly reproducible.

requirements.txt
reverendbedford: Good idea, and good catch on the other dependencies.

python/examples/README.md
reverendbedford: Added these things to the example specific requirements file.

reverendbedford:
> My final comment:
> 
> I think torch.manual_seed doesn't guarantee reproducibility in different environments. The SVI example has two stochastic aspects: training and inference. If you save the checkpoint and reload it in the notebook, the trained model should be reproducible. But then IDK how to make the inference part (i.e., sampling the posterior) truly reproducible.

Actually I think the SVI examples in pyzag are fully-reproducible by setting the random seed in `torch`.  But we'll see in a minute.

hugary1995 (approved):

PR #188, opened 2024-12-17
Generalize elastic constant conversion

### LINKED ISSUE #182: LinearCombination does not work with nonlinear parameter
coefficients in LinearCombination do not take nonlinear parameters.

hugary1995:
- Separate out conversion functions for elastic constants so that they could be reused elsewhere independent of the models
- Fix cross-referencing when declare nonlinear parameter -- same change applied to BufferStore
- Separate out generic converters for better code reuse: see `ElasticityConverter.h`
- Do not calculate derivatives if the input elastic constants are not nonlinear parameters
- Move all elasticity-related models into the `elasticity` subfolder


include/neml2/models/solid_mechanics/elasticity/ElasticityTensor.h
```diff
@@ -0,0 +1,188 @@
...
+
+namespace neml2
+{
+template <class Derived, std::size_t N>
+OptionSet
+ElasticityTensor<Derived, N>::expected_options()
+{
+  OptionSet options = Derived::expected_options();
+
+  MultiEnumSelection type_selection({"LAME_FIRST_CONSTANT",
```
reverendbedford: Honestly `"LAME_LAMBDA"` may be better.  This always makes me have to think about which is the first parameter...

include/neml2/models/solid_mechanics/elasticity/ElasticityConverter.h
```diff
@@ -0,0 +1,120 @@
...
+      input_values[i] = *input[i];
+    return convert(input_values);
+  }
+
+private:
+  /// Generate ordering of the input parameters
+  void assert_ascending(const ConverterKey & ps) const
+  {
+    for (std::size_t i = 1; i < N; ++i)
+      neml_assert(
```
reverendbedford:
```suggestion
      neml_assert_debug(
```

I assume, as this is an internal error.

reverendbedford (changes requested):

include/neml2/models/solid_mechanics/elasticity/ElasticityConverter.h
hugary1995: This is internal. But I think it doesn't hurt to make this check a non `_dbg` one, because it's only called in the constructor and doesn't affect run speed.

hugary1995:
Don't merge yet, one last change tomorrow morning.

include/neml2/models/solid_mechanics/elasticity/ElasticityConverter.h
reverendbedford: Okay.

reverendbedford (approved):

PR #187, opened 2024-12-17
Support spatial velocity gradient input for crystal models

reverendbedford:
Add some trivial models and tensor types for splitting an `R2` into symmetric and skew parts.

Test a crystal model taking the spatial velocity gradient as input.

PR #192, opened 2024-12-19
Fix assign_output_derivative bug

hugary1995:
The `std::map::insert` does not replace the value if a key already exists. This prevents IFT from updating derivatives if the variable derivative cache is not cleared after _every_ forward evaluation.


hugary1995:
close in favor of #187 which is based on this

PR #195, opened 2024-12-23
JIT

hugary1995:
First step at getting JIT to work. While I say this is first step, it's mostly working with just a few caveats
1. JIT in ComposedModel is temporarily disabled (not due to any technical difficulty)
2. JIT in ImplicitUpdate is disabled.
3. JIT in TorchScriptFlowRate is disabled (this is something I need to look into).

Other than that, all models default to `jit = true`.

This PR also temporarily sits on top of #187. I'll rebase once #187 gets merged.


src/neml2/models/solid_mechanics/Normality.cxx
```diff
@@ -79,9 +79,9 @@ Normality::set_value(bool out, bool dout_din, bool d2out_din2)
   {
     SolvingNonlinearSystem guard(false);
     if (out && !dout_din)
-      _model.dvalue();
```
reverendbedford, line 82: Probably want to leave a comment explaining these.

src/neml2/models/VariableStore.cxx
```diff
@@ -181,4 +289,84 @@ VariableStore::collect_output_second_derivatives() const
...
+  }
+
+  if (dout)
+  {
+    for (const auto & yvar : yvars)
+    {
+      const auto & derivs = output_variable(yvar).derivatives();
+      for (const auto & xvar : xvars)
+      {
+        const auto & deriv = derivs.find(xvar);
```
reverendbedford, line 329: This initially bothered me, but I guess `find` is log(n) or something in the number of keys.  So no changes required, just making sure you thought about the cost (and similar for d2out).

src/neml2/models/Model.cxx
```diff
@@ -199,14 +225,132 @@ Model::zero_output()
...
+        /*force_outplace=*/false));
+    auto new_function =
+        std::make_unique<torch::jit::GraphFunction>(name() + ".forward",
+                                                    trace->graph,
+                                                    /*function_creator=*/nullptr,
+                                                    torch::jit::ExecutorExecutionMode::PROFILING);
+    traced_functions[forward_op_idx].emplace(new_schema, std::move(new_function));
+
+    // Rerun this method -- this time using the jitted graph (without tracing)
+    forward_maybe_jit(out, dout, d2out);
```
reverendbedford, line 308: Is it necessary to rerun?  It seems like it is, but surely the results are calculated along with the trace, or else how would they be checking?  I can't find anything in the API to give you the (checked) results along with the traced code.

reverendbedford (changes requested):
In addition to the minor comments, we need to add a unit test for turning JIT off and comparing results for JIT/non-JIT for a non-trivial model.

src/neml2/models/solid_mechanics/Normality.cxx
hugary1995: This is actually a very good catch...

src/neml2/models/Model.cxx
hugary1995: Me neither...

src/neml2/models/VariableStore.cxx
hugary1995: I am also not satisfied with the cost. The root cause of this is my poor choice of container for variable derivatives. I plan to fix that (along with every consequent mistake) in the future.

reverendbedford (approved):

PR #200, opened 2025-01-04
Speed up AD

hugary1995:


src/neml2/misc/math.cxx
```diff
@@ -196,45 +196,74 @@ skew_to_full(const Tensor & skew, Size dim)
...
   const auto opt = y.options().requires_grad(false);
 
   // Flatten y to handle arbitrarily shaped output
-  const auto yf = y.base_flatten().batch_expand(batch_sizes);
-  const auto G = Tensor::identity(yf.base_size(0), opt).batch_expand(yf.batch_sizes());
+  const auto yf = y.base_flatten();
+  const auto G = Scalar::full(1.0, opt).batch_expand(yf.batch_sizes());
 
-  auto dyf_dx = Tensor::zeros(batch_sizes, utils::add_shapes(yf.base_size(0), x.base_sizes()), opt);
+  // Initialie derivatives to zero
```
reverendbedford:
```suggestion
  // Initialize derivatives to zero
```

reverendbedford (changes requested):
Incredibly minor comment.

I'm not sure why this is faster though...

hugary1995:
> I'm not sure why this is faster though...

The speed up comes from three places
1. In math.cxx `jacrev`, we used to create a big identity grad tensor `G` to query Jacobian row-by-row. That's not necessary and has been replaced by a scalar.
2. In `Model::extract_AD_derivatives`, we used to take derivatives for all output variables w.r.t. all input variables (whose AD derivatives are requested). That's not necessary during the implicit solve. Now they are guarded by `is_dependent`. So for example, in the example `TorchScriptFlowRate`, equivalent plastic strain rate depends on von mises stress and temperature, and the derivative of equivalent plastic strain rate w.r.t. temperature is skipped during the implicit update.
3. When we take derivative w.r.t. multiple input arguments, we used to trace them one by one in a for-loop. Although we have retain_graph set to true, torch somehow decides to throw away the graph when the input argument has changed, i.e., when tracing dy/dx1, dy/dx2, dy/dx3, the graph is created three times. In this PR, I first gather all input arguments before firing up autograd, so one graph is created per output variable.

reverendbedford (approved):

PR #203, opened 2025-02-02
Remove linalg.h

hugary1995:
torch version 2.6.0 removed linalg.h

We need to get rid of it for compatibility reasons


hugary1995:
I'm bypassing review requirement because there isn't any actual change.

PR #206, opened 2025-02-05
Reorganize folder structure and speed up build

hugary1995:
- Reorganize and split libneml2 into submodules
- Forward declare wherever possible
- Improved the dianogstic interface
- Replace CrossRef<T> with TensorName
- Replace Storage.h with plain std::map<K, std::unique_ptr<V>>
- Define one math function per header
- Cleanup c10 and at namespaces
- Replace torch::tensor with native factory methods
- Alias at::Tensor to ATensor
- Fix template class instantiation on mac arm64
- Build and install 3rd party libraries at configure time
- Better configure and build preset support
- Register runner tests using CTest
- Update installation guide


cmake/Modules/DownloadDoxygen.cmake
```diff
@@ -0,0 +1,21 @@
...
+# Right now this only downloads the Doxygen binary for Linux.
+#
+# Influential variables:
+# - DOXYGEN_VERSION
+#
+# Output variables are those defined by FetchContent such as:
+# - doxygen_POPULATED
+# - doxygen_SOURCE_DIR
+
+if(UNIX AND NOT APPLE)
```
reverendbedford, line 12: Maybe throw a warning or something if not.

cmake/Modules/DownloadTorch.cmake
reverendbedford, line 1: Is this path to getting torch actually tested automatically?

CMakeLists.txt
```diff
@@ -75,40 +62,75 @@ endif()
...
 # ----------------------------------------------------------------------------
-# Subdirectories
+# For relocatable install
+# ----------------------------------------------------------------------------
+if(UNIX AND APPLE)
+      set(EXEC_DIR "@loader_path")
+elseif(UNIX AND NOT APPLE)
+      set(EXEC_DIR "$ORIGIN")
+else()
+      install(CODE "message(WARNING \"The NEML2 installation may not be relocatable, and LD_LIBRARY_PATH shall be used to dynamically load libraries\")")
```
reverendbedford: Probably "should" rather than "shall"

include/neml2/misc/types.h
```diff
@@ -24,88 +24,76 @@
...
-  using std::variant<Size, torch::Tensor>::variant;
-
-  /// @return a pointer to the torch::Tensor representing the traceable size if it is traceable, otherwise a nullptr
-  const torch::Tensor * traceable() const noexcept;
-
-  /// @return the concrete size (without any traceable information)
-  Size concrete() const;
+using ATensor = at::Tensor;
+
+/// Fixed width dtypes (mirroring the definition in <torch/csrc/api/include/torch/types.h>)
```
reverendbedford, line 44: What are we gaining by having this huge, static list of aliases?  Also: pretty sure this is the main reason 2.3 no longer works... I'd consider something more portable.

doc/content/installation/integration.md
```diff
@@ -0,0 +1,49 @@
...
+- `neml2_dispatcher`
+- `neml2_driver`
+- `neml2_jit`
+- `neml2_misc`
+- `neml2_model`
+- `neml2_solver`
+- `neml2_tensor`
+- `neml2_user_tensor`
+
+The names of the libraries are self-explanatory. As a general rule of thumb, the library names collide with the header hierarchy. That is, if your code includes `#include "neml2/tensors/SR2.h"`, then your program shall be linked against the corresponding tensor library using `-lneml2_tensor`, etc.
```
reverendbedford: Probably "should" instead of "shall"

reverendbedford (changes requested):
1. Document torch version compatibility.
2. See my comment in a header file somewhere, I suspect that's the reason why you broke backward compatibility with a lot of old version.s
3. Doesn't really matter, but why are we referencing the `at` namespace directly now?
4. Something you've done broke the notebook python training examples...  If you get stuck I can help figure it out.  You need to run them on the GPU to find the bug.  (Yes, yes, we should run these in the test suite).

I can't remember what else you asked me to do to review, I'll check with you tomorrow.  I'll also try compiling on my mac.

cmake/Modules/DownloadDoxygen.cmake
hugary1995: There's a warning in doc/CMakeLists.txt.

cmake/Modules/DownloadTorch.cmake
hugary1995: Not really -- other than my local testing. I think this will be part of our expanded test matrix once we get gh enterprise.

include/neml2/misc/types.h
hugary1995: We gain compile speed by not including <torch/types.h> (which is pretty dumb and includes almost everything from torch). I've removed the offending device list, so we should be good now.

hugary1995:
> 1. Document torch version compatibility.
> 2. See my comment in a header file somewhere, I suspect that's the reason why you broke backward compatibility with a lot of old version.s
> 3. Doesn't really matter, but why are we referencing the `at` namespace directly now?
> 4. Something you've done broke the notebook python training examples...  If you get stuck I can help figure it out.  You need to run them on the GPU to find the bug.  (Yes, yes, we should run these in the test suite).
> 
> I can't remember what else you asked me to do to review, I'll check with you tomorrow. I'll also try compiling on my mac.

I'd like to defer 1 until after we have enterprise runners. I don't want to put something down now just with my local testing.
2 should be fixed now.
3 is because I'm not include <torch/type.h> anywhere to trade for compile speed.
4 I'll look...

hugary1995:
In my defense, the notebook failure (with cuda) is already in main...

reverendbedford (approved):

PR #211, opened 2025-02-12
Integrate work dispatchers into our drivers

reverendbedford:
Let users optionally provide a work dispatcher to handle running work through a driver.  Default option will still be to pass everything into the model once.


hugary1995 (approved):
I had some really minor comments so I just pushed a few commits to address them. I've also addressed most of my own comments during the rebasing, so technically speaking I'm not an independent reviewer. But nevertheless the changes you made look good to me.

PR #216, opened 2025-02-23
Newest benchmarks with dispatchers

hugary1995:
- Support parsing devices from input file
- Duplicate a factory for each thread, while input options and global settings are shared across all threads.
- Schedulers now default capacity to batch size
- Change how dispatcher and scheduler coordinate: The dispatcher and the scheduler use the main thread to coordinate work dispatching, and the dispatcher (in async mode) creates one thread per device to perform the work. The dispatcher manages the thread pool.


include/neml2/dispatchers/WorkDispatcher.h
```diff
@@ -53,8 +59,7 @@ namespace neml2
  * general, each work dispatch involves four steps:
  * 1. Work generation: The work generator generates the next \p n batches of work.
  * 2. Preprocessing: The dispatcher preprocesses the work.
- * 3. Do work: The dispatcher dispatches the work to a worker, and the worker completes and returns
- *    the result.
+ * 3. Do work: The worker performs the work.
  * 4. Postprocessing: The dispatcher postprocesses the result.
  *
```
reverendbedford, line 68: Let's add a brief summary of the threading model and its current limitations here (i.e. right now each device has its own thread).  Probably should leave some warning to the future about the stupid jit issues as well.

reverendbedford (changes requested):
We're going to need a better test suite for testing the dispatchers at some point, but I guess there's no point in doing that until we have gpu runners.  Anyway, just add a high level explanation of the threading model in the dispatcher comments and we'll call it good.

reverendbedford (approved):

PR #217, opened 2025-02-23
Remove nonlinear parameters

hugary1995:
The nonlinear parameter wasn't a perfect design. While it allows for models to use nonlinear parameters (models) as parameters, it fails to unify nonlinear parameter and regular models.

To address this design issue, I am replacing nonlinear parameter with a more flexible tensor cross referencing mechanism.

In a nutshell, we now support more ways of resolving a TensorName declared a parameter/buffer:
1. plain numeric literal
2. object name under the [Tensors] section
3. variable specifier

The first two already exist. The third one is new, and is only enabled when variable coupling is enabled for the parameter (using allow_nonlinar).

A variable specifier can take one of three forms:
1. `<model-name>.<variable-name>`
2. `<model-name>`
3. `<variable-name>`

The first form is the most precise, it specifies an output variable of a model. The value of that variable will be used as the parameter value.

The second form can be considered as a short-hand notation of the first form. It assumes that the model has one and only one output variable, and uses that variable as the parameter. If the model has zero or more than one output variable, this form is ambiguous and an error will be thrown.

The third form declares the parameter as an input variable. An InputParameter object will be created on-the-fly.


hugary1995:
This is stacked on top of #216

reverendbedford (approved):
You know subject to the tests passing...

hugary1995:
Since your approval, I
- moved the tensor name resolution function into ParameterStore to fix the linker error on mac
- fixed the github test reporting issue
- fixed recursive nonlinear parameter registration

The tests are all pasing, and since the new changes are mostly cosmetic, I'm merging this PR now.

PR #221, opened 2025-03-05
Tutorials

hugary1995:
* Implement a tutorial generation system to prevent tutorials from regression
* Add three set of tutorials: Models, Tensors, and Extension
* Miscellaneous library changes to facilitate tutorials


doc/tutorials/tutorials.md
```diff
@@ -0,0 +1,96 @@
...
+## Guides
+
+@htmlonly
+
+<div class="cards">
+  <div class="card">
+    <a href="tutorials-getting-started.html">
+    <div class="card_container">
+      <h2>Getting Started</h2>
+      <p>Get started as a user of NEML2: Choose the appropriate frontend that best suites your needs.</p>
```
reverendbedford:
```suggestion
      <p>Get started as a user of NEML2.  Choose the appropriate frontend that best suites your needs.</p>
```

doc/tutorials/models/evaluation_device/main.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Specifying a device
+
+A device is uniquely identified by a type, which specifies the type of machine it is (e.g. CPU or CUDA GPU), and a device index or ordinal, which identifies the specific compute device when there is more than one of a certain type. The device index is optional, and in its defaulted state represents (abstractly) "the current device". Further, there are two constraints on the value of the device index, if one is explicitly specified:
+- A negative index represents the current device, a non-negative index represents a specific, concrete device.
+- When the device type is CPU, the device index must be zero.
+
+In NEML2, the device can be specified using either a string or or a predefined device type.
+
+The string shall follow the schema `(cpu|cuda)[:<device-index>]`, where `cpu` or `cuda` specifies the device type, and `:<device-index>` optionally specifies a device index. For example, "cpu" represents the CPU and "cuda:1" specifies CUDA with device ID 1.
```
reverendbedford:
```suggestion
The string follows the schema `(cpu|cuda)[:<device-index>]`, where `cpu` or `cuda` specifies the device type, and `:<device-index>` optionally specifies a device index. For example, "cpu" represents the CPU and "cuda:1" specifies CUDA with device ID 1.
```

doc/tutorials/getting_started.md
```diff
@@ -8,11 +8,7 @@ Once the NEML2 library is built following the [installation guide](@ref install)
 
 ## Downstream C++ project
 
-NEML2 core capabilities are implemented as a _library_, not a _program_. As a library, it shall be used in another C++ program to parse and evaluate a material model defined in an [input file](@ref input-file). Boilerplate for the C++ program can be found [here](@ref cpp-backend), and external project integration is documented in the [installation guide](@ref external-project-integration).
-
-Related readings:
-- [Input file](@ref input-file) explains how NEML2 interacts with text-based input files to construct material models.
-- [C++ backend user guide](@ref cpp-backend) provides a basic example of using NEML2 in a C++ project.
+NEML2 core capabilities are implemented as a _library_, not a _program_. As a library, it shall be used in another C++ program to parse and evaluate a material model defined in an [input file](@ref tutorials-models-input-file). Boilerplate for the C++ program can be found in this [set of tutorials](#tutorials-models), and external project integration is documented in the [installation guide](@ref external-project-integration).
```
reverendbedford:
```suggestion
NEML2 core capabilities are implemented as a _library_, not a _program_. As a library, it will be used by another C++ program to parse and evaluate a material model defined in an [input file](@ref tutorials-models-input-file). Boilerplate for the C++ program can be found in this [set of tutorials](#tutorials-models), and external project integration is documented in the [installation guide](@ref external-project-integration).
```

doc/tutorials/contributing.md
```diff
@@ -101,3 +103,66 @@ To run a specific test case, use
...
+- Forces \f$\mathcal{F}\f$ (axis name `forces`): Variables defining the _external_ forces that drive the response of the material.
+- Old state \f$\mathcal{S}_n\f$ (axis name `old_state`): The state variables _prior to_ the current material update. In the time-discrete setting, these are the state variables from the previous time step.
+- Old forces \f$\mathcal{F}_n\f$ (axis name `old_forces`): The external forces _prior to_ the current material update. In the time-discrete setting, these are the forces from the previous time step.
+- Residual \f$\mathcal{R}\f$ (axis name `residual`): The residual defines an _implicit_ model/function. An implicit model is updated by solving for the state variables that result in zero residual.
+- Parameters \f$\mathcal{P}\f$ (axis name `parameters`): The (nonlinear) parameters.
+
+\note
+When authoring C++ source code, it is recommended to avoid hard-coding reserved axis names as pure strings. Instead, inlined `const` string names (defined in `neml2/models/LabeledAxisAccessor.h`) shall be used wherever possible, they are `STATE`, `OLD_STATE`, `FORCES`, `OLD_FORCES`, `RESIDUAL`, and `PARAMETERS` whose names are self-explanatory.
```
reverendbedford:
```suggestion
When authoring C++ source code, it is recommended to avoid hard-coding reserved axis names as pure strings. Instead, inlined `const` string names (defined in `neml2/models/LabeledAxisAccessor.h`) should be used wherever possible, they are `STATE`, `OLD_STATE`, `FORCES`, `OLD_FORCES`, `RESIDUAL`, and `PARAMETERS` whose names are self-explanatory.
```

doc/tutorials/contributing.md
```diff
@@ -101,3 +103,66 @@ To run a specific test case, use
...
+firefox build/dev/doc/build/html/index.html
+```
+
+## Code formatting and linting
+
+The C++ source code is formatted using `clang-format`. A `.clang-format` file is provided at the repository root specifying the formatting requirements. When using an IDE providing plugins or extensions to format C++ source code, it is important to
+1. Point the plugin/extension to use the `.clang-format` file located at NEML2's repository root.
+2. Associate file extensions `.h` and `.cxx` with C++.
+
+The Python scripts shall be formatted using `black`. Formatting requirements are specified under the `[black]` section in `pyproject.toml`.
```
reverendbedford:
```suggestion
The Python scripts must be formatted using `black`. Formatting requirements are specified under the `[black]` section in `pyproject.toml`.
```

doc/tutorials/contributing.md
```diff
@@ -1,8 +1,10 @@
-# Testing {#testing}
+@insert-title:tutorials-contributing
 
 [TOC]
 
-It is of paramount importance to ensure the correctness of the implementation. NEML2 comes with extensive testing for both its C++ backend and the Python package.
+## Model development
+
+Although NEML2 comes with a large collection modular building blocks for composing material models, it is sometimes necessary to write your own material models (and integrate them with existing NEML2 models). The [extension](#tutorials-extension) tutorial set demonstrates how a custom model can be implemented within the NEML2 framework and provides in-depth explanation for every step throughout the development process.
```
reverendbedford:
```suggestion
Although NEML2 comes with a large collection modular building blocks for composing material models, it is sometimes necessary to write your own material models (and integrate them with existing NEML2 models). The [extension](#tutorials-extension) tutorial set demonstrates how a custom model can be implemented within the NEML2 framework and provides an in-depth explanation for each step in the development process.
```

doc/tutorials/contributing.md
```diff
@@ -101,3 +103,66 @@ To run a specific test case, use
...
+
+All pull requests will be run through `clang-format` and `black` to ensure formatting consistency.
+
+For C++ linting, a `.clang-tidy` file is provided at the repository root to specify expected checks. Python linting is not currently enforced.
+
+## Naming conventions
+
+### Reserved axis names
+
+Recall that NEML2 models operates on _labeled tensors_, and that the collection of labels (with their corresponding layout) is called an labeled axis ([LabeledAxis](@ref neml2::LabeledAxis)). NEML2 predefines 6 sub-axes to categorize all the input, output and intermediate variables:
```
reverendbedford:
```suggestion
Recall that NEML2 models operates on _labeled tensors_, and that the collection of labels (with their corresponding layout) is called a labeled axis ([LabeledAxis](@ref neml2::LabeledAxis)). NEML2 predefines six sub-axes to categorize all the input, output and intermediate variables:
```

doc/tutorials/contributing.md
```diff
@@ -101,3 +103,66 @@ To run a specific test case, use
...
+- `;`: input file parsing ambiguity
+- `.`: clash with PyTorch parameter/buffer naming convention
+- `/`: separator reserved for nested variable name
+
+In the input file, the separator `/` is used to denote nested variable names. For example, `A/B/foo` specifies a variable named "foo" defined on the sub-axis named "B" which is a nested sub-axis of "A".
+
+### Source code naming conventions
+
+In NEML2 source code, the following naming conventions are recommended:
+- User-facing variables and option names should be _as descriptive as possible_. For example, the equivalent plastic strain is named "equivalent_plastic_strain". Note that white spaces, quotes, and left slashes are not allowed in the names. Underscores are recommended as an replacement for white spaces.
```
reverendbedford:
```suggestion
- User-facing variables and option names should be _as descriptive as possible_. For example, the equivalent plastic strain is named "equivalent_plastic_strain". Note that white spaces, quotes, and left slashes are not allowed in the names. Underscores are recommended as a replacement for white spaces.
```

doc/tutorials/contributing.md
```diff
@@ -101,3 +103,66 @@ To run a specific test case, use
...
+- Developer-facing variables and option names should use simple alphanumeric symbols. For example, the equivalent plastic strain is named "ep" in consistency with most of the existing literature.
+- Developner-facing member variables and option names should use the same alphanumeric symbols. For example, the member variable for the equivalent plastic strain is named `ep`. However, if the member variable is protected or private, it is recommended to prefix it with an underscore, i.e. `_ep`.
```
reverendbedford:
```suggestion
- Developer-facing member variables and option names should use the same alphanumeric symbols. For example, the member variable for the equivalent plastic strain is named `ep`. However, if the member variable is protected or private, it is recommended to prefix it with an underscore, i.e. `_ep`.
```

doc/tutorials/getting_started.md
```diff
@@ -42,13 +38,6 @@ Optional arguments:
...
-- [Input file](@ref input-file) explains how NEML2 interacts with text-based input files to construct material models.
-
 ## Python script
 
-The other option to evaluate NEML2 material models is to use the [NEML2 Pyton package](@ref python-package). As mentioned in the [installation guide](@ref install), NEML2 also provides an _experimental_ Python package which provides bindings for the primitive tensors and parsers for deserializing and running material models.
-
-Related readings:
-- [Input file](@ref input-file) explains how NEML2 interacts with text-based input files to construct material models.
-- [Python package user guide](@ref python-package) provides a basic example of using neml2 in a Python script.
+The other option to evaluate NEML2 material models is to use the NEML2 Pyton package. As mentioned in the [installation guide](@ref install), NEML2 also provides an _experimental_ Python package which provides bindings for the primitive tensors and parsers for deserializing and running material models. This [set of tutorials](#tutorials-models) describe the usage of the package.
```
reverendbedford:
```suggestion
The other option to evaluate NEML2 material models is to use the NEML2 Python package. As mentioned in the [installation guide](@ref install), NEML2 also provides an _experimental_ Python package which provides bindings for the primitive tensors and parsers for deserializing and running material models. This [set of tutorials](#tutorials-models) describes the usage of the package.
```

doc/tutorials/models/models.md
```diff
@@ -0,0 +1,10 @@
+@insert-title:tutorials-models
+
+NEML2 comes with a large collection of material models. These material models can be either used on their own or serve as building blocks for more complicated material models.
+
+Once NEML2 is [installed](@ref install), there is no need to write additional C++ code to define the material model of interest -- NEML2 can read from text-based [input files](@ref tutorials-models-input-file) to dynamically construct material models.
+
+This set of tutorials walk through typical use cases for interacting with NEML2 material models.
```
reverendbedford:
```suggestion
This set of tutorials walk through typical use cases for interacting with existing NEML2 material models.
```

doc/tutorials/models/input_file.md
```diff
@@ -55,50 +55,4 @@ On the other hand, other commonly used boolean flags such as "on"/"off", "1"/"0"
 
 **Tensor shape**: Shape of a tensor can also be parsed from a string. The string must start with "(" and end with ")". An array of comma-separated integers must be enclosed by the parentheses. For example, the string "(5,6,7)" can be parsed into a shape tuple of value `(5, 6, 7)`. Note that white spaces are not allowed between the parentheses and could lead to undefined behavior. An empty array, i.e. "()", however, is allowed and fully supported.
```
reverendbedford:
```suggestion
**Tensor shape**: The shape of a tensor can also be parsed from a string. The string must start with "(" and end with ")". An array of comma-separated integers must be enclosed by the parentheses. For example, the string "(5,6,7)" can be parsed into a shape tuple of value `(5, 6, 7)`. Note that white spaces are not allowed between the parentheses and could lead to undefined behavior. An empty array, i.e. "()", however, is allowed and fully supported.
```

doc/tutorials/models/running_your_first_model/main.md
```diff
@@ -0,0 +1,187 @@
...
+All NEML2 models, including this simple elasticity model under consideration, take the following general form
+\f[
+  y = f(x; p, b),
+\f]
+where \f$ x \f$ and \f$ y \f$ are respectively sets of input and output variables, \f$ p \f$ is the set of parameters, and \f$ b \f$ is the set of buffers. The utilities of parameters and buffers will be discussed in another [tutorial](#tutorials-models-model-parameters). The forward operator \f$ f \f$ is responsible for mapping from input variables \f$ x \f$ to \f$ y \f$. NEML2 provides three forward operators for all models:
+- neml2::Model::value calculates the output variables, i.e., \f$ y = f(x; p, b) \f$.
+- neml2::Model::dvalue calculates the derivatives of the output variables with respect to the input variables, i.e., \f$ \pdv{y}{x} = \pdv{f(x; p, b)}{x} \f$.
+- neml2::Model::value_and_dvalue calculates both the output variables and their derivatives.
+
+All three forward operators take a dictionary of variable values as input and return the requested output variables and/or their derivatives.
```
reverendbedford:
```suggestion
All three forward operators take a map/dictionary of variable values as input and return the requested output variables and/or their derivatives.
```

doc/tutorials/models/model_parameters/main.md
```diff
@@ -0,0 +1,202 @@
...
+\f]
+Pattern matching suggests the following set definitions:
+\f[
+  x = \left\{ \boldsymbol{\varepsilon} \right\}, \quad y = \left\{ \boldsymbol{\sigma} \right\}, \quad p = \left\{ K, G \right\}, \quad b = \varnothing.
+\f]
+
+## Parameter v.s. buffer
+
+Both \f$ K \f$ and \f$ G \f$ are here categorized as model parameters. The major differences between parameters and buffers are
+- Parameters are "trainable", whereas buffers are not. NEML2 can use automatic differentiation to calculate the derivative of output variables w.r.t. the model parameters.
```
reverendbedford:
```suggestion
- Parameters are "trainable", whereas buffers are not. NEML2 can use automatic differentiation to calculate the derivative of output variables with respect to the model parameters, but not for the model buffers.
```

doc/tutorials/models/model_parameters/main.md
```diff
@@ -0,0 +1,202 @@
...
+- Parameters can be (recursively) defined by other models, whereas buffers cannot.
```
reverendbedford: You may want to clarify this with an example.

doc/tutorials/models/model_parameters/main.md
```diff
@@ -0,0 +1,202 @@
...
+
+In summary, parameter is a more powerful superset of buffer.
```
reverendbedford:
```suggestion
In summary, a parameter is a more powerful superset of a buffer.  However, there is overhead cost associated with maintaining a parameter that buffers avoid.
```

doc/tutorials/models/model_parameters/main.md
```diff
@@ -0,0 +1,202 @@
...
+  Output:
+  ```
+  @attach-output:src4
+  ```
+
+</div>
+
+## Updating the parameter value
+
+Model parameters can always be changed by changing the input file. However, in certain cases (e.g., training and optimization), the parameter values should preferrably be updated at runtime (e.g., after each Epoch or optimization iteration).
```
reverendbedford:
```suggestion
Model parameters can always be changed by changing the input file. However, in certain cases (e.g., training and optimization), the parameter values should preferrably be updated at runtime (e.g., after each epoch or optimization iteration).
```

doc/tutorials/models/model_parameters/main.md
```diff
@@ -0,0 +1,202 @@
...
+Also recall that all NEML2 models can be written in the following general form
+\f[
+  y = f(x; p, b).
+\f]
+Pattern matching suggests the following set definitions:
+\f[
+  x = \left\{ \boldsymbol{\varepsilon} \right\}, \quad y = \left\{ \boldsymbol{\sigma} \right\}, \quad p = \left\{ K, G \right\}, \quad b = \varnothing.
+\f]
+
+## Parameter v.s. buffer
```
reverendbedford:
```suggestion
## Parameter vs buffer
```
though v.s. is actually more proper it looks weird.

doc/tutorials/models/evaluation_device/main.md
```diff
@@ -0,0 +1,103 @@
+@insert-title:tutorials-models-evaluation-device
+
+[TOC]
+
+## Device
+
+NEML2 inherits the definition of *device* from PyTorch. It is an abstraction representing the physical device where data is allocated and operated. NEML2 primarily concerns with two types of devices: CPU and CUDA.
```
reverendbedford:
```suggestion
NEML2 inherits the definition of *device* from PyTorch. It is an abstraction representing the physical device where data is stored and models executed. NEML2 is primarily concerned with two types of devices: CPU and CUDA.
```

doc/tutorials/models/evaluation_device/main.md
```diff
@@ -0,0 +1,103 @@
...
+- **CPU** stands for Central Processing Unit, and it tranditionally handles a wide range of scientific computing tasks.
+- **CUDA** functions as a programming platform that allows a GPU (Graphics Processing Unit) to act as a coprocessor to the CPU to handle specific computing tasks with massive parallelization.
+
+NEML2 offers several high-level mechanisms for users to strategically interact with CPU and CUDA.
+
+## Specifying a device
+
+A device is uniquely identified by a type, which specifies the type of machine it is (e.g. CPU or CUDA GPU), and a device index or ordinal, which identifies the specific compute device when there is more than one of a certain type. The device index is optional, and in its defaulted state represents (abstractly) "the current device". Further, there are two constraints on the value of the device index, if one is explicitly specified:
```
reverendbedford:
```suggestion
A device is uniquely identified by a type, which specifies the type of machine it is (e.g. CPU or CUDA GPU), and a device index or ordinal, which identifies the specific compute device when there is more than one of a certain type. The device index is optional, and in its default state represents (abstractly) "the current device". Further, there are two constraints on the value of the device index, if one is explicitly specified:
```

doc/tutorials/models/vectorization/main.md
```diff
@@ -0,0 +1,164 @@
...
+
+While the benefit of vectorization depends on the instruction support, instruction throughput, and register size, among other factors, it is generally expected that vectorization can provide a considerable amount of speedup to applicable operations: For example, given a device's instruction set with 1024-bit registers. These can be used to operate on 16 operands of 64-bits apiece. More specifically, if we are dealing with double-precision floating point numbers, a single instruction can do 16 operations at once. In other words, ideally, \f$ N \f$ operations can be completed using \f$ N/16 \f$ instructions. And again, ideally, this leads to 16 times faster execution than operations without vectorization.
+
+Although the implementation varies, both CPU and CUDA supports vectorization. NEML2 supports vectorization of model evaluation on both CPU and CUDA: All NEML2 models ensure that the data being processed, e.g., variables, parameters, buffers, are contiguously allocated, and that the same set of operations are applied to those contiguous chunks of data.
+
+## For loop v.s. vectorization: A numerical experiment
```
reverendbedford:
```suggestion
## For loop versus vectorization: A numerical experiment
```

doc/tutorials/models/vectorization/main.md
```diff
@@ -0,0 +1,164 @@
+@insert-title:tutorials-models-vectorization
+
+[TOC]
+
+## What is vectorization?
+
+More generally speaking, *array programming* is the computer science terminology vectorization is referring to. Both SIMD (single instruction, multiple data) operation and SIMT (single instruction, multiple threads) operation are implementation of array programming. These terms all refer to the solutions that allow the application of operations to an entire set of values at once.
```
reverendbedford:
```suggestion
Vectorization refers to *array programming*.  Both SIMD (single instruction, multiple data) operation and SIMT (single instruction, multiple threads) operation are implementations of array programming. These terms all refer to the methods that apply operations to an entire set of values at once.
```

doc/tutorials/models/vectorization/main.md
```diff
@@ -0,0 +1,164 @@
...
+
+While the benefit of vectorization depends on the instruction support, instruction throughput, and register size, among other factors, it is generally expected that vectorization can provide a considerable amount of speedup to applicable operations: For example, given a device's instruction set with 1024-bit registers. These can be used to operate on 16 operands of 64-bits apiece. More specifically, if we are dealing with double-precision floating point numbers, a single instruction can do 16 operations at once. In other words, ideally, \f$ N \f$ operations can be completed using \f$ N/16 \f$ instructions. And again, ideally, this leads to 16 times faster execution than operations without vectorization.
```
reverendbedford:
```suggestion
While the benefit of vectorization depends on the instruction support, instruction throughput, and register size, among other factors, it is generally expected that vectorization can provide a considerable amount of speedup to many operations.  For example, given a device's instruction set with 1024-bit registers. These can be used to operate on 16 operands of 64-bits apiece. More specifically, if we are dealing with double-precision floating point numbers, a single instruction can do 16 operations at once. In other words, ideally, \f$ N \f$ operations can be completed using \f$ N/16 \f$ instructions. And again, ideally, this leads to 16 times faster execution than operations without vectorization.
```

doc/tutorials/models/vectorization/main.md
```diff
@@ -0,0 +1,164 @@
...
+
+Although the implementation varies, both CPU and CUDA supports vectorization. NEML2 supports vectorization of model evaluation on both CPU and CUDA: All NEML2 models ensure that the data being processed, e.g., variables, parameters, buffers, are contiguously allocated, and that the same set of operations are applied to those contiguous chunks of data.
```
reverendbedford:
```suggestion
Although the implementation varies, both CPU and CUDA supports vectorization. NEML2 supports vectorization of model evaluation on both CPU and CUDA.  All NEML2 models ensure that the data being processed, e.g., variables, parameters, buffers, are contiguously allocated, and that the same set of operations are applied to those contiguous chunks of data.
```

doc/tutorials/models/vectorization/main.md
```diff
@@ -0,0 +1,164 @@
...
+
+There are two ways of completing this task:
+- using a for loop, or
+- using vectorization
+
+### For loop
+
+The following code snippet shows how to use a for loop to perform the \f$ N \f$ evaluations. For convenience, the \f$ N \f$ strains are evenly spaced.
+
+<div class="tabbed">
```
reverendbedford, line 25: While I understand why you don't want to check in and run/regress this code is there a better way to just include it without having to write out the html in the markdown?

doc/tutorials/models/vectorization/main.md
```diff
@@ -0,0 +1,164 @@
...
+| CPU    | 10     | 0.061    | 0.003         |
+|        | 1000   | 0.232    | 0.004         |
+|        | 100000 | 16.782   | 0.041         |
+| CUDA   | 10     | 0.145    | 0.004         |
+|        | 1000   | 0.337    | 0.004         |
+|        | 100000 | 26.176   | 0.005         |
+
+## Remarks
+
+While there are many other factors contributing to the performance difference between the for loop approach and the vectorization approach, we hope this simple experiment can serve as a gentle introduction to the benefit of vectorization.
```
reverendbedford:
```suggestion
While there are many other factors contributing to the performance difference between the for loop approach and the vectorization approach, we hope this simple experiment serves as a reasonable introduction to the benefit of vectorization.
```

doc/tutorials/models/model_composition/main.md
```diff
@@ -0,0 +1,340 @@
+@insert-title:tutorials-models-model-composition
+
+[TOC]
+
+## Problem description
+
+We have been working with the linear, isotropic elasticity model in the previosu tutorials. We started with that example because it is arguably the simplest possible material model in the context of solid mechanics. It is simple not just because of the simplicity in the description of the material behavior, but also due to the fact that its mathematical formulation only involves one equation.
```
reverendbedford:
```suggestion
We have been working with the linear, isotropic elasticity model in the previous tutorials. We started with that example because it is arguably the simplest possible material model in the context of solid mechanics. It is simple not just because of the simplicity in the description of the material behavior, but also due to the fact that its mathematical formulation only involves one linear equation.
```

doc/tutorials/models/model_composition/main.md
```diff
@@ -0,0 +1,340 @@
...
+
```
reverendbedford, line 10: You need some transition sentence here.  Something like 

> NEML2 can run and optimize much more complicated, nonlinear models.

doc/tutorials/models/model_composition/main.md
```diff
@@ -0,0 +1,340 @@
...
+  ```
+  @endsource
+
+  Output:
+  ```
+  @attach-output:src2
+  ```
+
+</div>
+
```
reverendbedford, line 164: There is mystery blank space in the html rendering here for some reason.  If you can't figure out why don't worry, I don't care that much about it.

doc/tutorials/models/model_composition/main.md
```diff
@@ -0,0 +1,340 @@
...
+## Running the models: the easy way
+
+We were able to successfully calculate \f$ \dot{\boldsymbol{b}} \f$ by
+1. calculating \f$ \bar{a} \f$ by evaluating \f$ \eqref{1} \f$,
+2. calculating \f$ \bar{b} \f$ by evaluating \f$ \eqref{2} \f$,
+3. setting the two coefficients of  \f$ \eqref{3} \f$ to be \f$ \bar{b} \f$ and \f$ \bar{a} \f$ respectively,
+4. calculating \f$ \dot{\boldsymbol{b}} \f$ by evaluating \f$ \eqref{3} \f$.
+
+However, that is not ideal because we had to
+- manually evaluate the equations and figuring out the evaluation order, and
```
reverendbedford:
```suggestion
- Manually evaluate the equations and figure out the evaluation order, and
```

doc/tutorials/models/model_composition/main.md
```diff
@@ -0,0 +1,340 @@
...
+- manually set the parameters in \f$ \eqref{3} \f$ as outputs from \f$ \eqref{1}\&\eqref{2} \f$.
```
reverendbedford:
```suggestion
- Manually set the parameters in \f$ \eqref{3} \f$ as outputs from \f$ \eqref{1}\&\eqref{2} \f$.
```

doc/tutorials/models/model_composition/main.md
```diff
@@ -0,0 +1,340 @@
...
+
+Moreover, such method is not scalable when the number of equations, variables, and parameters increase.
```
reverendbedford:
```suggestion
This manual method is not scalable when the number of equations, variables, and parameters increase.
```

doc/tutorials/models/model_composition/main.md
```diff
@@ -0,0 +1,340 @@
...
+
+  Output:
+  ```
+  @attach-output:src4
+  ```
+
+</div>
+
+Note that the composed model "eq" automatically
+- identified the input variables \f$ \boldsymbol{a} \f$ and \f$ \boldsymbol{b} \f$,
```
reverendbedford:
```suggestion
- Identified the input variables \f$ \boldsymbol{a} \f$ and \f$ \boldsymbol{b} \f$,
```

doc/tutorials/models/model_composition/main.md
```diff
@@ -0,0 +1,340 @@
...
+  @endsource
+
+  Output:
+  ```
+  @attach-output:src4
+  ```
+
+</div>
+
+Note that the composed model "eq" automatically
```
reverendbedford:
```suggestion
Note that the composed model "eq" automatically:
```

doc/tutorials/models/model_composition/main.md
```diff
@@ -0,0 +1,340 @@
...
+- identified the input variables \f$ \boldsymbol{a} \f$ and \f$ \boldsymbol{b} \f$,
+- identified the output variable \f$ \dot{\boldsymbol{b}} \f$,
```
reverendbedford:
```suggestion
- Identified the output variable \f$ \dot{\boldsymbol{b}} \f$,
```

doc/tutorials/models/model_composition/main.md
```diff
@@ -0,0 +1,340 @@
...
+- registered parameters of \f$ \eqref{3} \f$ as input variables, and
```
reverendbedford:
```suggestion
- Registered the parameters of \f$ \eqref{3} \f$ as input variables, and
```

doc/tutorials/models/model_composition/main.md
```diff
@@ -0,0 +1,340 @@
...
+- sorted out the evaluation order.
```
reverendbedford:
```suggestion
- Sorted out the evaluation order.
```

doc/tutorials/models/model_parameters_revisited/main.md
```diff
@@ -0,0 +1,424 @@
+@insert-title:tutorials-models-model-parameters-revisited
+
+[TOC]
+
+## Problem description
+
+In a [previous tutorial](#tutorials-models-model-parameters), we briefly explained the usage of model parameters and their retrieval and update. In this tutorial, we will introduce other ways of specifying model parameters in the input file and how they facilitate model composition.
+
+Consider the following equations describing the stress-strain relation accounting for a thermal eigenstrain:
```
reverendbedford, line 9:
```suggestion
Consider the following equations describing an elastic stress-strain relation accounting for a thermal eigenstrain:
```

doc/tutorials/models/model_parameters_revisited/main.md
```diff
@@ -0,0 +1,424 @@
...
+  @endsource
+
+  Output:
+  ```
+  @attach-output:src6
+  ```
+
+</div>
+
+When the referenced model has more than one output variables, the variable specification becomes ambiguous. A more precise variable specifier in the form of `<model-name>.<variable-name>` can be used to remove such ambiguity. For example, although unncessary (as [ScalarLinearInterpolation](#scalarlinearinterpolation) defines one and only one output variable), the above example input is equivalent to
```
reverendbedford, line 296:
```suggestion
When the referenced model has more than one output variable, the variable specification becomes ambiguous. A more precise variable specifier in the form of `<model-name>.<variable-name>` can be used to remove such ambiguity. For example, although unnecessary (as [ScalarLinearInterpolation](#scalarlinearinterpolation) defines one and only one output variable), the above example input is equivalent to
```

doc/tutorials/models/model_parameters_revisited/main.md
```diff
@@ -0,0 +1,424 @@
...
+    coefficient_types = 'BULK_MODULUS SHEAR_MODULUS'
+    coefficients = 'K G'
+  []
+  [eq]
+    type = ComposedModel
+    models = 'eq1 eq2 eq3'
+  []
+[]
+```
+Again, NEML2 automatically reflects such change in model structure:
```
reverendbedford, line 383:
```suggestion
Again, NEML2 automatically reflects such changes in model structure:
```

doc/tutorials/models/implicit_model/main.md
```diff
@@ -0,0 +1,264 @@
+@insert-title:tutorials-models-implicit-model
+
+[TOC]
+
+## Problem description
+
+One of the most notable differences between constitutive models and feed-forward neural networks is that implicitly updating certain stiff systems is much more computationally efficient than explicit algorithms.
```
reverendbedford:
```suggestion
One of the most notable differences between constitutive models and feed-forward neural networks is that updating certain stiff systems with implicit methods is often more computationally efficient compared to explicit algorithms.
```

doc/tutorials/models/implicit_model/main.md
```diff
@@ -0,0 +1,264 @@
...
+## Solving the system of equations
+
+Once the system of equations are properly defined, we can use the [ImplicitUpdate](#implicitupdate) to solve the system of equations. The [ImplicitUpdate](#implicitupdate) is responsible for the following:
+- (Re)declare the solution to the system of equations as output variables.
+- Validate the shape of input variables and residual variables to make sure the system is square.
+- Assemble the residual vector and Jacobian matrix of the underlying linear system.
+- Invoke a *solver* to solve the system of equations.
+- Apply the implicit function theorem to calculate exact derivatives (up to machine precision).
+
+At the time of writing, NEML2 offers three fully vectorized Newton solvers to be used in conjunction with [ImplicitUpdate](#implicitupdate):
```
reverendbedford:
```suggestion
NEML2 offers three fully vectorized Newton solvers to be used in conjunction with [ImplicitUpdate](#implicitupdate):
```

We'll just have to keep current :)

doc/tutorials/models/implicit_model/main.md
```diff
@@ -0,0 +1,264 @@
...
+- [Newton](#newton), the vanilla version of the Newton-Raphson algorithm which always takes the "full" step.
```
reverendbedford:
```suggestion
- [Newton](#newton), the (vectorized) vanilla version of the Newton-Raphson algorithm which always takes the "full" step.
```

doc/tutorials/models/implicit_model/main.md
```diff
@@ -0,0 +1,264 @@
...
+
+</div>
+
+## Remarks on the implicit function theorem
+
+Unlike other regular models, declaring variables on the *correct* sub-axes is important because NEML2 relies on the reserved sub-axes (`state`, `forces`, etc.)
+- To determine whether the variable value and derivatives should be calculated during the assembly of residual and Jacobian. For example, the derivatives with respect to all variables on the `forces` sub-axis are skipped because they are not required in the assembly of the linear system.
+- To efficiently reuse the factorization of the system Jacobian when applying the implicit function theorem.
+
+As long as models are defined using the *correct* sub-axis definitions and satisfy some mild continuity requirements, **NEML2 guarantees the correctness of the outcoming variable derivatives** after one or more implicit updates, up to machine precision. The same guarantee also applies to user-defined custom models.
```
reverendbedford:
```suggestion
As long as models are defined using the *correct* sub-axis definitions and satisfy some mild continuity requirements, **NEML2 guarantees the correctness of the variable derivatives** after one or more implicit updates, up to machine precision. The same guarantee also applies to user-defined custom models.
```

doc/tutorials/models/implicit_model/main.md
```diff
@@ -0,0 +1,264 @@
...
+
+This is a significant advantage compared to some of the alternative constitutive model libraries, especially in the context of coupling with external PDE solvers. For example, in the context of finite element method, thermodynamic forces (e.g. strain) are calculated at each quadrature point, and the constitutive library (e.g. NEML2) is responsible for updating the thermodynamic state variables (e.g. stress, plastic strain, etc.), which are then used in the residual definition of the discretized PDE. Therefore, the exact derivatives of the state variables w.r.t. the forces are the key to the assembly of the exact Jacobian of the descretized PDE, which is in turn the fundamental requirement for optimal convergence for many nonlinear solvers.
```
reverendbedford:
```suggestion
This is a significant advantage compared to some of the alternative constitutive model libraries, especially in the context of coupling with external PDE solvers. For example, in the context of finite element method, thermodynamic forces (e.g. strain) are calculated at each quadrature point, and the constitutive library (e.g. NEML2) is responsible for updating the thermodynamic state variables (e.g. stress, plastic strain, etc.), which are then used in the residual definition of the discretized PDE. Therefore, the exact derivatives of the state variables with respect to the forces are the key to the assembly of the exact Jacobian of the descretized PDE, which is in turn the fundamental requirement for optimal convergence for many nonlinear solvers.
```

doc/tutorials/models/transient_driver/main.md
```diff
@@ -0,0 +1,126 @@
+@insert-title:tutorials-models-transient-driver
+
+[TOC]
+
+## Problem description
+
+In the previous tutorial, we have demonstrated the use of [ImplicitUpdate](#implicitupdate) to perform constitutive update by solving an implicit system of equations. In particular, we have addressed the constitutive update in the following form
```
reverendbedford:
```suggestion
In the previous tutorial, we demonstrated the use of [ImplicitUpdate](#implicitupdate) to perform constitutive update by solving an implicit system of equations. In particular, we have addressed the constitutive update in the following form
```

doc/tutorials/models/transient_driver/main.md
```diff
@@ -0,0 +1,126 @@
...
+\f{align}
+  \mathbf{s}_{n+1} = f(\mathbf{f}_{n+1}, \mathbf{s}_n, \mathbf{f}_n; \mathbf{p}). \label{form1}
+\f}
+In other words, the constitutive update takes a recursive form: Given the state and external force of the system (\f$\mathbf{s}_n\f$ and \f$\mathbf{f}_n\f$) at time step \f$n\f$, as well as the "forces" \f$\mathbf{f}_{n+1}\f$ driving the system to advance to the next step \f$n+1\f$, the model yields the state of the system for the next step \f$\mathbf{s}_{n+1}\f$.
+
+This form of constitutive update is oftentimes sufficient when coupling with external PDE solvers: The PDE solver calculates the external driving force \f$\mathbf{f}_{n+1}\f$ and asks NEML2 to advance the state of system \f$\mathbf{s}_{n+1}\f$.
+
+However, in many other applications such as parameter calibration, it is favorable to let NEML2 drive the constitutive update (recursively) to effectively simulate the transient response of the material. The corresponding initial-value problem can be formally written as: Given initial conditions \f$\mathbf{s}_0\f$ and \f$\mathbf{f}_0\f$, find, \f$\forall n \in [0, N-1]\f$,
```
reverendbedford:
```suggestion
However, in many other applications such as parameter calibration, it is favorable to let NEML2 drive the constitutive update (recursively) to effectively simulate the transient response of the material. The corresponding initial-value problem can be formally written as: 

Given initial conditions \f$\mathbf{s}_0\f$ and \f$\mathbf{f}_0\f$, find, \f$\forall n \in [0, N-1]\f$,
```

doc/tutorials/models/transient_driver/main.md
```diff
@@ -0,0 +1,126 @@
...
+\f{align}
+  \mathbf{s}_{n+1} = f(\mathbf{f}_{n+1}, \mathbf{s}_n, \mathbf{f}_n; \mathbf{p}), \label{form2}
+\f}
+where \f$N\f$ is the total number of steps used to discretize the "transient".
+
+NEML2 models, including those composed from others, can only be used to describe the first form of constitutive \f$\eqref{form1}\f$, i.e., a single step constitutive update. In order to simulate the initial-value problem defined in \f$\eqref{form2}\f$, a neml2::TransientDriver can be used to perform the recursive constitutive update to obtain the transient.
```
reverendbedford:
```suggestion
NEML2 models, including those composed from submodels, can only be used to describe the first form of constitutive \f$\eqref{form1}\f$, i.e., a single step constitutive update. In order to simulate the initial-value problem defined in \f$\eqref{form2}\f$, a neml2::TransientDriver can be used to perform the recursive constitutive update to obtain the transient response.
```

doc/tutorials/models/transient_driver/main.md
```diff
@@ -0,0 +1,126 @@
...
+{
+  load_input("input.i");
+  auto & driver = get_driver("driver");
+  driver.run();
+}
+```
+@endsource
+
+\note
+The Python binding for neml2::Driver does not yet exist. Instead, a separate Python package named pyzag can be used in conjunction with NEML2 to perform recursive constitutive update, in a much more efficient manner.
```
reverendbedford:
```suggestion
The Python binding for neml2::Driver does not yet exist. Instead, a separate Python package named pyzag can be used in conjunction with NEML2 to perform recursive constitutive updates in a much more efficient manner compared to the NEML2 drivers, which loop over each point in the requested time series sequentially.
```

doc/tutorials/models/just_in_time_compilation/main.md
```diff
@@ -0,0 +1,310 @@
+@insert-title:tutorials-models-just-in-time-compilation
+
+[TOC]
+
+## Problem description
+
+In NEML2, all tensor operations are traceable. The trace of an operation records the operator type, a stack of arguments and outputs, together with additional context. Multiple operations performed in sequence can be traced together into a *graph* representing the flow of data through operators. Such graph representation is primarily used for two purposes:
+- just-in-time (JIT) compilation and optimization of the operations
```
reverendbedford:
```suggestion
- Just-in-time (JIT) compilation and optimization of the operations
```

doc/tutorials/models/just_in_time_compilation/main.md
```diff
@@ -0,0 +1,310 @@
...
+- backward automatic-differentiation (AD)
```
reverendbedford:
```suggestion
- Backward automatic-differentiation (AD)
```

doc/tutorials/models/just_in_time_compilation/main.md
```diff
@@ -0,0 +1,310 @@
...
+
+This tutorial illustrates the utility of JIT compilation of NEML2 models, and the a [later tutorial](#tutorials-optimization-automatic-differentiation) demonstrates the use of backward AD to calculate parameter derivatives.
```
reverendbedford:
```suggestion
This tutorial illustrates the utility of JIT compilation of NEML2 models, and a [later tutorial](#tutorials-optimization-automatic-differentiation) demonstrates the use of backward AD to calculate parameter derivatives.
```

doc/tutorials/models/just_in_time_compilation/main.md
```diff
@@ -0,0 +1,310 @@
...
+  @attach-output:src6
+  ```
+
+</div>
+
+Note how the optimized graph successfully identifies the common subexpression \f$ t - t_n \f$ and reuses it in all three equations.
+
+## Limitations
+
+JIT optimization and compilation isn't the holy grail for improving performance of all models. For tensor operations that branch based on variable data, the traced graph cannot capture such data dependency and would potentially produce wrong results. For models defining the derivatives relying on backward automatic differentiation, NEML2 is not able to generate efficient traced graphs.
```
reverendbedford:
```suggestion
JIT optimization and compilation isn't the holy grail for improving performance of all models. For tensor operations that branch based on variable data, the traced graph cannot capture such data dependency and would potentially produce wrong results. NEML2 is unable to generate traced graphs for models that include derivatives of other models in the forward evaluation when those derivatives are defined with automatic differentiation.
```

doc/tutorials/models/just_in_time_compilation/main.md
```diff
@@ -0,0 +1,310 @@
...
+
+Due to these limitations, certain models disable the use of JIT compilation. The most notable case is [ImplicitUpdate](#implicitupdate) due to its use of Newton-Raphson solvers which are in general data dependent.
```
reverendbedford:
```suggestion
Due to these limitations, certain models disable the use of JIT compilation. The most notable case is [ImplicitUpdate](#implicitupdate) due to its use of Newton-Raphson solvers which are in general data dependent.  However, the portions of the complete model defining the implicit function to solve can often benefit from JIT compilation.
```

reverendbedford (commented):
1. We can save this for later, but there has to be a better way to input source code and output versus having to write the html tags directly in the markdown.

Review in progress...

doc/tutorials/tensors/tensors.md
```diff
@@ -0,0 +1,7 @@
+@insert-title:tutorials-tensors
+
+NEML2 is equipped with its own tensor library. Currently, PyTorch (more specifically, ATen) is the only supported tensor backend in NEML2. Therefore, all tensor types in NEML2 directly inherit from `torch::Tensor`. In the future, support for other tensor backends may be added, but the public-facing interfaces will remain largely the same.
+
+This set of tutorials introduce basic API usage of NEML2's tensor library:
```
reverendbedford:
```suggestion
This set of tutorials introduce the basic API of NEML2's tensor library:
```

doc/tutorials/tensors/tensor_types/main.md
```diff
@@ -0,0 +1,97 @@
+@insert-title:tutorials-tensors-tensor-types
+
+[TOC]
+
+NEML2 tensors are extremely similar to the tensor type provided by its default backend ATen. In fact, all NEML2 tensors inherit from `at::Tensor`, *the* tensor type from ATen. What's different from ATen is that the NEML2 tensor library provides the following enhancements:
+- Explicit distinction between *batch* and base dimensions
```
reverendbedford:
```suggestion
- Explicit distinction between *batch* and *base* dimensions
```

doc/tutorials/tensors/tensor_types/main.md
```diff
@@ -0,0 +1,97 @@
...
+we could assign it a batch dimension of 3, and the resulting shape of the batched tensor becomes
+\f[
+  (3, 100, 5; 13, 2).
+\f]
+Similarly, if the batch dimension is 0 or 5, the resulting shape would become
+\f{align*}
+  (; 3, 100, 5, 13, 2), \\
+  (3, 100, 5, 13, 2;).
+\f}
+It is valid for the \f$;\f$ delimiter to appear at the beginning or the end of the shape tuple.
```
reverendbedford:
```suggestion
It is valid for the \f$;\f$ delimiter to appear at the beginning or the end of the shape tuple.  If it appears at the beginning it means that tensor has no batch dimensions.  If it appears at the end it means the tensor is a batched scalar.
```

I can't remember is the last statement true?

doc/tutorials/tensors/tensor_types/main.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Why batching?
+
+The definition of batching is fairly straightforward yet arbitrary. A fair question to ask is "why another layer of abstraction on top of the regular tensor shape definition"? While it is difficult to give a thorough list of technical reasons at this point of the tutorial, let us briefly cover the high-level motivation behind this design choice.
```
reverendbedford:
```suggestion
The definition of batching is fairly straightforward. However, fair question to ask is "why another layer of abstraction on top of the regular tensor shape definition"? Let's briefly consider the high-level motivation behind this design choice.
```

doc/tutorials/tensors/tensor_types/main.md
```diff
@@ -0,0 +1,97 @@
...
+
+**Scalar broadcasting**: With more details to be covered in a [later tutorial](#tutorials-tensors-broadcasting) on broadcasting, explicit batching allows general broadcasting rules between scalar and other primitive tensor types.
+
+**Shape generalization**: As mentioned in the [tutorial](#tutorials-models-just-in-time-compilation) on JIT compilation, NEML2 supports tracing tensor operations into graphs. The PyTorch tracing engine does not support tracing operations on tensor shapes, and so users are left with two options:
+- Trace the graph with example input variables having the same shape as real input variables to be used later on.
+- Re-trace the graph whenever any input variable has a different shape.
+
+The first option implies a potential memory bound at the stage of JIT compilation, and is difficult to guarantee in practice. The second option imposes a significant runtime overhead whenever re-tracing is deemed necessary.
```
reverendbedford:
```suggestion
The first option implies a potential memory bound at the stage of JIT compilation, and is difficult to guarantee in practice. The second option imposes a significant runtime overhead whenever re-tracing is necessary.
```

doc/tutorials/tensors/tensor_types/main.md
```diff
@@ -0,0 +1,97 @@
...
+| [WWR4](#neml2::WWR4)               | \f$(3,3)\f$             | Rank-4 tensor with skew symmetry                                 |
+| [R5](#neml2::R5)                   | \f$(3,3,3,3,3)\f$       | Rank-5 tensor                                                    |
+| [SSFR5](#neml2::SSFR5)             | \f$(6,6,3)\f$           | Rank-5 tensor with minor symmetry on base dimensions 0-3         |
+| [R8](#neml2::R8)                   | \f$(3,3,3,3,3,3,3,3)\f$ | Rank-8 tensor                                                    |
+| [SSSSR8](#neml2::SSSSR8)           | \f$(6,6,6,6)\f$         | Rank-8 tensor with minor symmetry                                |
+| [Rot](#neml2::Rot)                 | \f$(3)\f$               | Rotation tensor represented in the Rodrigues form                |
+| [Quaternion](#neml2::Quaternion)   | \f$(4)\f$               | Quaternion                                                       |
+| [MillerIndex](#neml2::MillerIndex) | \f$(3)\f$               | Crystal direction or lattice plane represented as Miller indices |
+
+Furthermore, all primitive tensor types can be declared as variables, parameters, and buffers in a model.
```
reverendbedford:
```suggestion
All primitive tensor types can be declared as variables, parameters, and buffers in a model.
```

doc/tutorials/tensors/tensor_creation/main.md
```diff
@@ -0,0 +1,117 @@
...
+  from neml2.tensors import SR2
+
+  A = SR2.zeros((5, 3), dtype=torch.float32, device=torch.device("cpu"))
+  ```
+  @endsource
+
+</div>
+
+\note
+All factory methods are listed [here](@ref tensor_creation).
```
reverendbedford:
```suggestion
All the factory methods are listed [here](@ref tensor_creation).
```

doc/tutorials/tensors/indexing/main.md
```diff
@@ -0,0 +1,372 @@
+@insert-title:tutorials-tensors-indexing
+
+[TOC]
+
+Related readings:
+- NumPy manual [Indexing on ndarrays](https://numpy.org/doc/2.2/user/basics.indexing.html)
+- PyTorch documentation [Tensor Indexing API](https://pytorch.org/cppdocs/notes/tensor_indexing.html)
+
+## Batch v.s. base indexing
```
reverendbedford:
```suggestion
## Batch versus base indexing
```

doc/tutorials/tensors/indexing/main.md
```diff
@@ -0,0 +1,372 @@
+@insert-title:tutorials-tensors-indexing
+
+[TOC]
+
+Related readings:
```
reverendbedford:
```suggestion
Related reading:
```

doc/tutorials/tensors/tensor_view/main.md
```diff
@@ -0,0 +1,106 @@
+@insert-title:tutorials-tensors-tensor-view
+
+[TOC]
+
+## View v.s. copy
```
reverendbedford:
```suggestion
## View versus copy
```

doc/tutorials/tensors/tensor_view/main.md
```diff
@@ -0,0 +1,106 @@
...
+
+NEML2 allows a tensor to be a view of an existing tensor. As the name suggests, a tensor view shares the same underlying data with the tensor it is viewing into. In other words, supporting tensor view avoids data copy. Moreover, tensor views optionally (and often) reinterpret the shape and/or striding of the original data, allowing for fast and memory efficient reshaping, slicing, and element-wise operations.
+
+\remark
+Tensor view is the enabling technique behind efficient assembly of implicit systems.
```
reverendbedford:
```suggestion
Tensor views enable efficient assembly of implicit systems.
```

doc/tutorials/tensors/tensor_view/main.md
```diff
@@ -0,0 +1,106 @@
...
+
+In fact, all indexing mechanisms covered in the [previous tutorial](#tutorials-tensors-indexing) are creating tensor views, i.e., zero copy, negligible allocation. In addition to those indexing API, NEML2 also provides flexible tensor reshaping API (documented in neml2::TensorBase).
+
+## Who touched my data?
+
+Tensor view avoids explicit data copy, which means
```
reverendbedford:
```suggestion
Tensor views avoid explicit data copying, which means:
```

doc/tutorials/tensors/broadcasting/main.md
```diff
@@ -0,0 +1,136 @@
...
+[TOC]
+
+Related reading: NumPy manual on [Broadcasting](https://numpy.org/doc/stable/user/basics.broadcasting.html)
+
+## NEML2's broadcasting rules
+
+NEML's broadcasting rules are built on top of NumPy's general broadcasting rules.
+
+When operating on two tensors, NEML2 compares each dimension's size pair one-by-one, from right to left. Before the comparison, the shapes of two tensors are aligned at the final batch dimension. The comparison starts with the trailing (i.e., rightmost) batch dimension and works its way left. Two batch dimensions are compatible (broadcastable) when
+- their sizes are equal,
```
reverendbedford:
```suggestion
- Their sizes are equal,
```

doc/tutorials/tensors/broadcasting/main.md
```diff
@@ -0,0 +1,136 @@
...
+- one of them has size one, or
```
reverendbedford:
```suggestion
- One of them has size one, or
```

doc/tutorials/tensors/broadcasting/main.md
```diff
@@ -0,0 +1,136 @@
...
+- one of them does not exist.
```
reverendbedford:
```suggestion
- One of them does not exist.
```

doc/tutorials/tensors/broadcasting/main.md
```diff
@@ -0,0 +1,136 @@
...
+
+If all dimensions are compatible, the two tensors are *batch-broadcastable*. Many operations in NEML2 require operands to be batch-broadcastable, and if not, an exception will be thrown (most of them only throw the exception in Debug mode).
```
reverendbedford:
```suggestion
If all dimensions are compatible, the two tensors are *batch-broadcastable*. Many operations in NEML2 require operands to be batch-broadcastable, and if not, an exception will be thrown (typically only in Debug mode).
```

reverendbedford (commented):
Next one.

doc/tutorials/extension/extension.md
```diff
@@ -0,0 +1,46 @@
+@insert-title:tutorials-extension
+
+[TOC]
+
+\note
+Before diving into the following tutorials, it is optional but recommended to read the [Contributing](#tutorials-contributing) guide.
```
reverendbedford:
```suggestion
Before diving into the following tutorials we recommend reading the [Contributing](#tutorials-contributing) guide.
```

doc/tutorials/extension/extension.md
```diff
@@ -0,0 +1,46 @@
...
+NEML2 models can be easily extended or adapted to model the material behavior of your interest.
+
+## Problem description
+
+This set of tutorials define a custom equation to model the conventional *projectile motion* problem:
+\f{align}
+  \dot{\boldsymbol{x}} & = \boldsymbol{v}, \label{1} \\
+  \dot{\boldsymbol{v}} & = \boldsymbol{a} = \boldsymbol{g} - \mu \boldsymbol{v}, \label{2}
+\f}
+where \f$\boldsymbol{x}\f$ and \f$\boldsymbol{v}\f$ are the position and velocity of the cannonball, respectively. \f$\boldsymbol{g}\f$ is the gravitational acceleration, and \f$\mu\f$ is the dynamic viscosity. The projectile's trajectory can be numerically integrated using the backward-Euler method:
```
reverendbedford:
```suggestion
where \f$\boldsymbol{x}\f$ and \f$\boldsymbol{v}\f$ are the position and velocity of the projectile, respectively. \f$\boldsymbol{g}\f$ is the gravitational acceleration, and \f$\mu\f$ is the dynamic viscosity. The projectile's trajectory can be numerically integrated using the backward-Euler method:
```

doc/tutorials/extension/extension.md
```diff
@@ -0,0 +1,46 @@
...
+    \boldsymbol{v}
+  \end{Bmatrix} & = \mathop{\mathrm{root}}\limits_{\boldsymbol{x}, \boldsymbol{v}} \left( \mathbf{r} \right), \label{4}
+\f}
+subject to appropriate initial conditions \f$\boldsymbol{x}_0\f$ and \f$\boldsymbol{v}_0\f$
+
+## Seaching for available models
+
+Among these equations:
+- \f$\eqref{1}\f$ and \f$\eqref{3}\f$ can be defined using [VecBackwardEulerTimeIntegration](#vecbackwardeulertimeintegration).
+- \f$\eqref{2}\f$ is not available in NEML2, and is the focus of this set of tutorials. We will name it as `ProjectileAcceleration`.
```
reverendbedford:
```suggestion
- \f$\eqref{2}\f$ is not available in NEML2, and is the focus of this set of tutorials. We will name it  `ProjectileAcceleration`.
```

doc/tutorials/extension/extension.md
reverendbedford, line 1: You don't have the navigation links at the bottom for this one.

doc/tutorials/extension/connection_to_input_file/main.md
reverendbedford, line 1: I suggest changing the title to "Connection to input files" or "Connection to the input file"

doc/tutorials/extension/connection_to_input_file/main.md
```diff
@@ -0,0 +1,167 @@
+@insert-title:tutorials-extension-connection-to-input-file
+
+[TOC]
+
+NEML2 utilizes the well-established [registry-factory pattern](https://en.wikipedia.org/wiki/Factory_method_pattern) to allow for runtime object creation and distributed class registration. In NEML2, neml2::Parser, neml2::Registry, and neml2::Factory work together to create objects defined in the input file.
+
+## Parser
+
+A class is said to be runtime-manufacturable if it's constructor can be retrieved at runtime using a string identification of its symbol. To allow for a unified object creation signature, runtime-manufacturable objects (e.g., models) are required to derive from neml2::NEML2Object and provide a constructor and a static method with the following signatures
```
reverendbedford:
```suggestion
A class is said to be runtime-manufacturable if its constructor can be retrieved at runtime using a string identification of its symbol. To allow for a unified object creation signature, runtime-manufacturable objects (e.g., models) are required to derive from neml2::NEML2Object and provide a constructor and a static method with the following signatures
```

doc/tutorials/extension/connection_to_input_file/main.md
```diff
@@ -0,0 +1,167 @@
...
+- The parser is responsible for bridging the gap between the expected input options and the user-specified options.
+
+The following steps take place for each object during input file parsing:
+1. Parser extracts object type from the reserved field `type`.
+2. Parser retrieves *expected* input options from the registry.
+3. Parser iterates over *user-specified* options and override default options.
+
+At the end of this process, a set of input options are gathered for the subsequent object creation.
+
+All of the gathered options are stored in the NEML2 factory so that objects can be created at runtime per user's request.
```
reverendbedford:
```suggestion
All of the gathered options are stored in the NEML2 factory so that objects can be created at runtime per the user's request.
```

doc/tutorials/extension/connection_to_input_file/main.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Input file
+
+Before defining the input file options for our custom model, it is always a good idea to first design the syntax. For example, in this equation we are mapping from the projectile's current velocity to its acceleration, therefore we'd like to allow user to specify the variable names corresponding to the velocity and acceleration. In addition, since the equation is parametrized by the gravitational acceleration \f$\boldsymbol{g}\f$ and the dynamic viscosity \f$\nu\f$, we should allow user to set their values directly inside the input file (without modifying the source code).
```
reverendbedford:
```suggestion
Before defining the input file options for our custom model, it is a good idea to first design the syntax. For example, in this equation we are mapping from the projectile's current velocity to its acceleration, therefore we'd like to allow user to specify the variable names corresponding to the velocity and acceleration. In addition, since the equation is parametrized by the gravitational acceleration \f$\boldsymbol{g}\f$ and the dynamic viscosity \f$\nu\f$, we should allow user to set their values directly inside the input file (without modifying the source code).
```

doc/tutorials/extension/connection_to_input_file/main.md
```diff
@@ -0,0 +1,167 @@
...
+  options.set<TensorName<Scalar>>("dynamic_viscosity");
+  return options;
+}
+}
+```
+@endsource
+
+## Registry
+
+The registry is a RAII-style singleton responsible for storing the mapping from the string identification of each runtime-manufacturable object to the constructor pointer along with the expected input file options.
```
reverendbedford:
```suggestion
The registry is a [RAII-style](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) singleton responsible for storing the mapping from the string identification of each runtime-manufacturable object to the constructor pointer along with the expected input file options.
```

doc/tutorials/extension/connection_to_input_file/main.md
```diff
@@ -0,0 +1,167 @@
...
+    // Variable and parameter declarations are to be added in the following tutorials
+{
+}
+}
+```
+@endsource
+
+The following tutorials will explain how to declare variables and parameters in the constructor.
+
+Suppose an input file named "inpu.i" has the following content
```
reverendbedford:
```suggestion
Suppose an input file named "input.i" has the following content
```

doc/tutorials/extension/argument_declaration/main.md
```diff
@@ -0,0 +1,99 @@
+@insert-title:tutorials-extension-argument-declaration
+
+[TOC]
+
+Recall that all NEML2 models can be written in the following form
```
reverendbedford:
```suggestion
All NEML2 models can be written in the following form
```

doc/tutorials/extension/model_composition/main.md
```diff
@@ -0,0 +1,183 @@
...
+Among these equations:
+- \f$\eqref{1}\f$ and \f$\eqref{3}\f$ can be defined using [VecBackwardEulerTimeIntegration](#vecbackwardeulertimeintegration).
+- \f$\eqref{2}\f$ is the custom model `ProjectileAcceleration` which we have implemented in previous tutorials.
+- \f$\eqref{4}\f$ is the [ImplicitUpdate](#implicitupdate).
+
+This tutorial demonstrates that our custom model `ProjectileAcceleration` can be composed with other existing, predefined NEML2 models.
+
+## Input file
+
+Our custom model `ProjectileAcceleration` can be composed with other existing, predefined NEML2 models. The following input file composes the constitutive model for a single-step update for 5 projectiles each with a different dynamic viscosity, i.e., the shape of "mu" is \f$(5;)\f$.
```
reverendbedford:
```suggestion
The following input file composes the constitutive model for a single-step update for 5 projectiles each with a different dynamic viscosity, i.e., the shape of "mu" is \f$(5;)\f$.
```

(You just said that in the previous paragraph)

doc/tutorials/extension/model_visualization/main.md
```diff
@@ -1,4 +1,8 @@
-# Model visualization {#model-visualization}
+@insert-title:tutorials-extension-model-visualization
+
+[TOC]
+
+## Motivation
 
 During the process of model development, it is of general interest (and is also very much encouraged) to reuse existing models. This is particularly useful and effective owing to the extreme modular design of NEML2, i.e., model composition.
```
reverendbedford:
```suggestion
Users should reuse existing NEML2 models as much as possible. This is particularly effective owing to the extremely modular design of NEML2, i.e., model composition.
```

reverendbedford (commented):
Here's the next one.

doc/tutorials/optimization/optimization.md
```diff
@@ -0,0 +1,17 @@
+@insert-title:tutorials-optimization
+
+The previous tutorials have illustrated the use of NEML2 constitutive models in the "feed-forward" setting, i.e., the model maps from input variables to output variables with a given parametrization, i.e.
```
reverendbedford:
```suggestion
The previous tutorials illustrated the use of NEML2 constitutive models in the "feed-forward" setting, i.e., the model maps from input variables to output variables with a given parametrization, i.e.
```

doc/tutorials/optimization/optimization.md
```diff
@@ -0,0 +1,17 @@
...
+\f]
+Recall that \f$p\f$ and \f$b\f$ are respectively the parameters and the buffers of the model.
+
+Another interesting use of NEML2 constitutive models is *parameter calibration*: With given input variables \f$x\f$, find the optimal parameter set \f$p^*\f$ such that
+\f[
+  p^* = \mathop{\mathrm{argmin}}\limits_{p} l(f(x; p, b)),
+\f]
+where \f$l\f$ is oftentimes referred to as the loss (or objective) function defining optimality.
+
+This set of tutorials demonstrate the use of PyTorch [Autograd](https://pytorch.org/tutorials/beginner/blitz/autograd_tutorial.html) to calculate parameter derivatives (\f$\pdv{l}{p}\f$) which is a necessary ingredient in all gradient-based optimizers.
```
reverendbedford:
```suggestion
This set of tutorials demonstrate the use of PyTorch [Autograd](https://pytorch.org/tutorials/beginner/blitz/autograd_tutorial.html) to calculate parameter derivatives (\f$\pdv{l}{p}\f$), which is a necessary ingredient in all gradient-based optimizers.
```

doc/tutorials/optimization/optimization.md
reverendbedford, line 1: No forward/backward links at the bottom again.

doc/tutorials/optimization/automatic_differentiation/main.md
```diff
@@ -0,0 +1,74 @@
+@insert-title:tutorials-optimization-automatic-differentiation
+
+[TOC]
+
+## Model definition
+
+For demonstration purposes, let us go back to our example -- the linear isotropic elasticity model:
```
reverendbedford:
```suggestion
For demonstration purposes, let us go back to an earlier example model -- linear isotropic elasticity:
```

doc/tutorials/optimization/automatic_differentiation/main.md
```diff
@@ -0,0 +1,74 @@
...
+@attach-output:src1
+```
+
+## Automatic differentiation
+
+NEML2 tensors can be used interchangeably with PyTorch tensors in a seamless fashion. Function graph traced through NEML2 tensor operations can be back-propagated using PyTorch's autograd engine.
+
+For example, the following Python script demonstrates the calculation of \f$\pdv{l}{p}\f$ using an arbitrary, scalar-valued loss function \f$l\f$ obtained by a combination of NEML2 operations and PyTorch operations.
+
+@source:src2
```
reverendbedford, line 41: Strictly speaking it's weird you go to only python at this point and don't have c++, but I understand why given the next section.

doc/tutorials/optimization/parameter_calibration/main.md
```diff
@@ -0,0 +1,85 @@
+@insert-title:tutorials-optimization-parameter-calibration
+
+[TOC]
+
+## Problem definition
+
+Suppose we are given the experimental data of one pair of strain \f$\boldsymbol{\varepsilon}^*\f$ and stress \f$\boldsymbol{\sigma}^*\f$ measurement of a specimen, and we are tasked to infer the bulk \f$K\f$ and shear \f$G\f$ modulus of the material (assuming it's homogeneous, isotropic, purely elastic, and small strain assumptions hold).
```
reverendbedford:
```suggestion
Suppose we are given the experimental data of one pair of strain \f$\boldsymbol{\varepsilon}^*\f$ and stress \f$\boldsymbol{\sigma}^*\f$ measurement of a specimen, and we are tasked to infer the bulk \f$K\f$ and shear \f$G\f$ modulus of the material.
```

reverendbedford (changes requested):
Last one.

I think we should add one more single page tutorial called "Next Steps" or similar.  It should refresh people on:
1. The Syntax, System, and Physics documentation.  Emphasize that we already have most solid mechanics models people might want.
2. Link (again) the contribution guide.
3. Mention pyzag and that training with those methods (adjoint, time-batching) is much more efficient for realistic models.  I don't want people complaining when they can't get good results with AD through transients...

Should be short, just a page.

doc/tutorials/extension/extension.md
hugary1995: The tutorial landing page can't have a navigation link, as it will just point to the next set of tutorials. Also, I mean it has a list of subsections for people to click on, so...

doc/tutorials/optimization/optimization.md
hugary1995: Same reply, I don't think landing pages should have navigation links...

doc/tutorials/models/model_parameters/main.md
hugary1995: I added a link here pointing to "Model parameters revisited" which should explain this.

doc/tutorials/models/vectorization/main.md
hugary1995: I think so, yes. Actually that may be a better approach than what I have put together so far. I created an issue #222 to keep track of this.

doc/tutorials/optimization/automatic_differentiation/main.md
hugary1995: Yeah, it's somewhat tricky because we can't use matplotlib in c++, so we'll have to split out the plotting part. After I get to #222 I'll revisit this.

reverendbedford (approved):

PR #223, opened 2025-03-07
Build wheels

hugary1995:
I have zero hope in this, but it doesn't hurt to try.


hugary1995:
okay this should be mostly working now, except for an embarrassing error:
```
Could not install packages due to an OSError: [Errno 28] No space left on device
```

I'll get back to this once we get the enterprise account.

reverendbedford (approved):
I looked and you have to have a paid teams or enterprise account to use their larger runners :(

PR #225, opened 2025-03-10
Allow derivmaps to return unbatched results

reverendbedford:
- Let the derivmap join routines return unbatch results if the contributions from the subbatches want that.
- Add a consistency check for debug mode.


hugary1995 (approved):
I'm okay with this but eventually when this is all ironed out we need tests and documentation.

PR #226, opened 2025-03-19
Miscellaneous changes to default settings

### LINKED ISSUE #208: Do not manual_seed
Some functions accept an optional argument to re-seed the rng seed.

This is undesirable as it potentially overrides the seed user sets outside.

### LINKED ISSUE #209: Context segregation
Currently the `Settings` object is created at parsing time, during which some default options such as dtype, num_threads, etc. are set.

This approach is undesirable because it makes context management depend on parsing.

The correct approach would be to store default options and create (and destroy) the corresponding context for each forward operator.

hugary1995:
* Remove interop_threads and intraop_threads as those are supposed to be set outside NEML2.
* Change the Settings constructor to a static method
* Factor out default settings into preprocessor definitions
* Remove the Settings singleton in Factory
* Move random seed from Rot and Orientation to Driver
* Respect torch default dtype
* Add thread sanitizer cmake presets and github workflow for detecting race conditions

close #209
close #208


src/neml2/drivers/Driver.cxx
```diff
@@ -44,6 +46,9 @@ Driver::expected_options()
   options.set<bool>("verbose") = false;
   options.set("verbose").doc() = "Whether to output additional logging information";
 
+  options.set<Size>("random_seed") = -1;
```
reverendbedford: Why not just the usual check to see if the user set it?  Why use the magic number?

reverendbedford (changes requested):
One minor suggestion.

src/neml2/drivers/Driver.cxx
hugary1995: Good point.

reverendbedford (approved):

PR #228, opened 2025-03-20
Expand dispatchers and improve documentation

reverendbedford:
- Add a simple MPI dispatcher
- Add hwloc as a dependency
- Switch timpi to a shared library
- Improve dispatcher documentation


src/neml2/CMakeLists.txt
```diff
@@ -64,6 +64,26 @@ if(NEML2_WORK_DISPATCHER)
 
   if(TIMPI_DOWNLOADED)
     install(DIRECTORY ${TIMPI_INCLUDE_DIR}/timpi TYPE INCLUDE COMPONENT libneml2)
+    install(TARGETS ${HWLOC_LIBRARY} TYPE LIBRARY COMPONENT libneml2)
```
hugary1995: Should this install go under `if(HWLOC_DOWNLOADED)`?

src/neml2/dispatchers/SimpleMPIScheduler.cxx
```diff
@@ -0,0 +1,165 @@
...
+  // - If multiple CUDA devices are present, make sure each CUDA device has a concrete
+  //   (nonnegative), unique device ID
+  bool has_multiple_cuda_devices = device_list.size() > 1 && cuda;
+  if (has_multiple_cuda_devices)
+  {
+    std::set<DeviceIndex> cuda_device_ids;
+    for (const auto & device : device_list)
+    {
+      if (device.is_cpu())
+        continue;
```
hugary1995:
```suggestion
```

src/neml2/dispatchers/SimpleMPIScheduler.cxx
```diff
@@ -0,0 +1,165 @@
...
+    _comm(TIMPI::Communicator(MPI_COMM_WORLD))
+{
+}
+
+void
+SimpleMPIScheduler::setup()
+{
+  WorkScheduler::setup();
+
+  const auto & device_list = input_options().get<std::vector<Device>>("devices");
```
hugary1995: this is just `_available_devices`

src/neml2/dispatchers/SimpleMPIScheduler.cxx
```diff
@@ -0,0 +1,165 @@
...
+{
+  OptionSet options = WorkScheduler::expected_options();
+  options.doc() =
+      "Dispatch work to a single device selected based on processor ID in given batch sizes.";
+
+  options.set<std::vector<Device>>("devices");
+  options.set("devices").doc() = "List of devices to dispatch work to";
+
+  options.set<std::size_t>("batch_size");
+  options.set("batch_size").doc() = "Batch size";
```
hugary1995: Should we allow different batch sizes for devices?

src/neml2/dispatchers/SimpleMPIScheduler.cxx
```diff
@@ -0,0 +1,165 @@
...
+
+  options.set<std::size_t>("capacity");
+  options.set("capacity").doc() =
+      "Maximum number of work items that can be dispatched, default to batch_size";
```
hugary1995: Same here

src/neml2/dispatchers/SimpleMPIScheduler.cxx
```diff
@@ -0,0 +1,165 @@
...
+  int id = static_cast<int>(hasher(hostname) % std::numeric_limits<int>::max());
+  hwloc_topology_destroy(topology);
+
+  // Make a new communicator based on this hashed hostname
+  TIMPI::Communicator new_comm;
+  _comm.split(id, _comm.rank(), new_comm);
+  // Assign our device index based on the new communicator
+  _device_index = new_comm.rank();
+  neml_assert(_device_index < _available_devices.size(),
+              "MPI split by host would require too many devices");
```
hugary1995: I understand `_device_index == _available_devices.size() - 1`.
I also understand why `_device_index > _available_devices.size() - 1` should error.

How about the other case `_device_index < _available_devices.size() - 1`? Does it mean one of the devices will remain idle?

hugary1995 (changes requested):

src/neml2/dispatchers/SimpleMPIScheduler.cxx
reverendbedford: I fixed the docstring, i'm fine with a device being idle.  Though maybe we should throw a warning?

src/neml2/dispatchers/SimpleMPIScheduler.cxx
reverendbedford: Yep

src/neml2/dispatchers/SimpleMPIScheduler.cxx
reverendbedford: Yep

reverendbedford:
And Laura complained I was up late working last night.

reverendbedford:
I assume you want to actually display your animation, but this is ready for review in terms of finishing everything I wanted to finish.

doc/content/system/scheduler.md
```diff
@@ -1,4 +1,68 @@
...
+
+\note
+Currently NEML2 only supports CPU and CUDA compute devices.  Additional accelerators supported by pytorch including Apple [mps](https://pytorch.org/docs/stable/mps.html) and Intel [xpu](https://pytorch.org/docs/stable/xpu.html) may also work, but are not officially supported.
+
+NEML2 manages work scheduling, dispatch, and joining with two types of objects: dispatchers and schedulers.  These objects divide up work along a single batch axis by subdividing that axis into smaller chunks (subbatches) to run on a given device.
+
+## Work dispatchers
+
+NEML2 currently only has a single type of work dispatcher.  The dispatcher maintains a main thread which schedules and distributes work according the algorithm provided by the selected work scheduler.  The main thread dispatches the work to a thread pool that maintains a specific thread for each device.  The thread picks up the task (representing a batch of work), sends the work to the device, and returns the work, once completed, to the main thread.
```
hugary1995: How about the `async=false` case? I think technically they are two different dispatchers, even though they are implemented by the same object in the same file.

doc/content/system/scheduler.md
```diff
@@ -1,4 +1,68 @@
...
+
+Setting up the work dispatcher requires providing it with lambda functions describing how to:
+1. Actually run a subbatch of work on a device
+2. Reduce the final collection of subbatches into a single result
+3. Preprocess a subbatch before running it (for example, send it to the device)
+4. Postprocess a subbatch before joining (for example, send it back to the cpu)
+5. Initialize the device thread.
```
hugary1995: Seems like it should be sub-batch:
https://books.google.com/ngrams/graph?content=subbatch%2Csub-batch&year_start=1940&year_end=2022&corpus=en&smoothing=3&case_insensitive=false

doc/content/system/scheduler.md
```diff
@@ -1,4 +1,68 @@
...
+
+NEML2 provides helper routines for assembling these lambdas for models returning the value of the model and/or the first derivatives of the results with respect to the inputs.
+
+## Work schedulers
+
+A work scheduler determines how to divide the full batch of model updates into smaller subbatches and assign those to different compute devices.  NEML2 provides several types of schedulers representing a combination of scheduling models:
+1. A single CPU task has exclusive access to a single compute device.
+2. A single CPU task has exclusive access to one or more compute devices.
+3. A set of CPU tasks shares access to one or more compute devices (this is work in progress and not available in the current release of NEML2).
```
hugary1995:
```suggestion
3. Multiple CPU tasks share access to one or more compute devices (this is work in progress and not available in the current release of NEML2).
```

doc/content/system/scheduler.md
```diff
@@ -1,4 +1,68 @@
...
+
+A CPU task here can either mean a single program calling NEML2, for example the runner program provided with the NEML2 release, or a single MPI rank in a distributed parallel application.  The third type of scheduler covers the most general heterogenous compute architecture, where a collection of local CPU threads have shared access to a collection of accelerator devices.  However, the first two types of schedulers can also be used in distributed computing environments, with the caveat that each task/MPI rank must have *exclusive* access to a collection of accelerators.
+
+### SimpleScheduler
+
+The [SimpleScheduler](#simplescheduler) is the simplest option.  It dispatches work from a single instance of NEML2 to a single device.  The user provides:
+1. The torch-compatible name of the target device, e.g. `"cuda:0"`.  This cannot be the CPU.
```
hugary1995: I thought this can use CPU.

doc/content/system/scheduler.md
```diff
@@ -1,4 +1,68 @@
...
+2. In cases where each NEML2 instance, for example each MPI rank, only has exclusive access to a single accelerator.  For example, running one MPI rank per node and each node has a single GPU.
+
+It also demonstrates the basic concepts applied in other schedulers.
+
+### SimpleMPIScheduler
+
+The [SimpleMPIScheduler](#simplempischeduler) is an extension of the [SimpleScheduler](#simplescheduler) which better distributes work in MPI parallel applications calling NEML2.  This scheduler assigns each MPI rank a unique accelerator device out of a list of devices provided by the user.  This MPI rank sends work to its assigned device exclusively, using the same algorithm for work scheduling as in the [SimpleScheduler](#simplescheduler).  The difference in inputs between the [SimpleMPIScheduler](#simplempischeduler) and the [SimpleScheduler](#simplescheduler) is just that the user now provides a list of devices, a list of subbatch sizes, and (optionally), a list of capacities.  Each item in the list is a single device available to MPI ranks.
```
hugary1995:
```suggestion
The [SimpleMPIScheduler](#simplempischeduler) is an extension of the [SimpleScheduler](#simplescheduler) which better distributes work in MPI parallel applications calling NEML2.  This scheduler assigns each MPI rank a unique accelerator device out of a list of devices provided by the user.  Each MPI rank sends work to its assigned device exclusively, using the same algorithm for work scheduling as in the [SimpleScheduler](#simplescheduler).  The difference in inputs between the [SimpleMPIScheduler](#simplempischeduler) and the [SimpleScheduler](#simplescheduler) is just that the user now provides a list of devices, a list of subbatch sizes, and (optionally), a list of capacities.  Each item in the list is a single device available to MPI ranks.
```

doc/content/system/scheduler.md
```diff
@@ -1,4 +1,68 @@
...
+
+This scheduler assigns a rank to a device by first forming a local MPI communicator over the MPI ranks running on a single compute node (determined by the hostname provided by the [hwloc library](https://www.open-mpi.org/projects/hwloc/)).  Each MPI rank in the communicator is assigned one device from the list provided by the user.  The user must provide at least as many devices as the local communicator has MPI ranks.  This approach limits the number of MPI ranks a user can launch *per node* to the number of available accelerators.  For example, if a compute node has one CPU with 16 threads and 8 GPU accelerators the host application can only run 8 MPI ranks per node.  The scheduler will report an error if the host application tries to run it with more local ranks than user supplied devices.  This often means the user will need to [pin ranks](https://hpc-wiki.info/hpc/Binding/Pinning) to ensure their MPI application assigns only the required number of ranks per node.
```
hugary1995: Now that I think more about it, if we only use hwloc to get the hostname, why not just use `MPI_Get_processor_name` provided by the MPI implementation plus an all-to-all?

doc/content/system/scheduler.md
```diff
@@ -1,4 +1,68 @@
...
+
+This scheduler is useful for larger parallel jobs, but may limit the overall performance of the application as often there are more CPU cores available for MPI ranks than their are accelerators on a node.
```
hugary1995:
```suggestion
This scheduler is useful for larger parallel jobs, but may limit the overall performance of the application as often there are more CPU cores available for MPI ranks than accelerators on a node.
```

doc/content/system/scheduler.md
```diff
@@ -1,4 +1,68 @@
...
+
+### StaticHybridScheduler
+
+The [StaticHybridScheduler](#statichybridscheduler) allows a single instance of NEML2 to distribute work to many compute devices at once.  The user provides a list of devices, subbatch sizes, capacities, and (optionally) priorities.  The scheduler will greedily assign work to fill up each device, using the priority if provided to determine in which order to dole out work.  The work dispatcher will then run this work in parallel on all devices via the thread pool discussed above.
+
+This scheduler currently provides the best throughput for cases where a single NEML2 instance has access to multiple devices.  Unlike the simple schedulers, the device list can include the CPU itself so that the CPU thread running NEML2 can do work at the same time the accelerators are working on their subbatches.  We suggest the user determine the optimal subbatch size for each device, for example using the [SimpleScheduler](#simplescheduler), prior to running large jobs with this scheduling algorithm.
```
hugary1995: > Unlike the simple schedulers, the device list can include the CPU itself so that the CPU thread running NEML2 can do work at the same time the accelerators are working on their subbatches.

Similar question as before -- I thought the simple scheduler can use CPU.

src/neml2/dispatchers/SimpleMPIScheduler.cxx
```diff
@@ -0,0 +1,166 @@
...
+      "Number of capacities must match the number of devices and the number of batch sizes.");
+}
+
+void
+SimpleMPIScheduler::setup()
+{
+  WorkScheduler::setup();
+
+  // First pass:
+  // - Prohibit any CPU
```
hugary1995, line 77: Same question, why prohibiting CPU?

src/neml2/dispatchers/SimpleMPIScheduler.cxx
```diff
@@ -0,0 +1,166 @@
...
+  hwloc_topology_load(topology);  // actual detection
+  machine = hwloc_get_root_obj(topology);
+  std::hash<std::string> hasher;
+  std::string hostname(hwloc_obj_get_info_by_name(machine, "HostName"));
+  int id = static_cast<int>(hasher(hostname) % std::numeric_limits<int>::max());
+  hwloc_topology_destroy(topology);
+
+  // Make a new communicator based on this hashed hostname
+  TIMPI::Communicator new_comm;
+  _comm.split(id, _comm.rank(), new_comm);
```
hugary1995: Seems like the splitting is based purely on `id` which is hashed from hostname. I think `MPI_Get_processor_name` can do the same job and is portable, at least for both MPICH and OpenMPI.

hugary1995 (changes requested):

doc/content/system/scheduler.md
hugary1995: I'm okay with getting rid of `async=false`, but then we would lose an easy, debugger friendly dispatcher...

reverendbedford: As discussed I'll change it to default `true` but still document both behaviors.

doc/content/system/scheduler.md
reverendbedford: You are right...

doc/content/system/scheduler.md
reverendbedford: It's a made up word but sure.

src/neml2/dispatchers/SimpleMPIScheduler.cxx
reverendbedford: We don't have to but you're better off not using a scheduler if you're using this one with the cpu.

doc/content/system/scheduler.md
reverendbedford: You are correct...

src/neml2/dispatchers/SimpleMPIScheduler.cxx
reverendbedford: Yep

hugary1995 (approved):
LGTM.

PR #230, opened 2025-03-26
Add an example for Drucker-Prager surface

hugary1995:
We can already do this, but it will be easier for users if we have a direct example.


tests/regression/solid_mechanics/viscoplasticity/drucker_prager/model.i
```diff
@@ -0,0 +1,168 @@
...
+    type = SR2Invariant
+    invariant_type = 'I1'
+    tensor = 'state/M'
+    invariant = 'state/i1'
+  []
+  [effective_stress]
+    type = ScalarLinearCombination
+    from_var = 'state/svm state/i1'
+    to_var = 'state/s'
+    coefficients = '1 0.5' # coefficients for Drucker-Prager, note that von Mises stress is sqrt(3 J2)
```
reverendbedford: Isn't beta usually negative?  Of course now you'll have to regold.

```suggestion
    coefficients = '1 -0.5' # coefficients for Drucker-Prager, note that von Mises stress is sqrt(3 J2)
```

reverendbedford (changes requested):
I mean it serves it's function, but in the interest of not potentially confusing people...

reverendbedford (approved):

PR #232, opened 2025-03-26
Install files describing version number and hash

hugary1995:


hugary1995:
Also decrease min cmake from 3.28 -> 3.26

PR #236, opened 2025-03-27
Fix FindTIMPI.cmake

hugary1995:
I am currently listing timpi with all possible build type variants as search candidates. This results in ambiguity when multiple timpi, e.g., libtimpi_opt and libtimpi_dbg, coexist.

The solution is to explicitly list the candidate with the expected suffix.


hugary1995:
Thanks again @roystgnr

reverendbedford (approved):

PR #239, opened 2025-03-29
Support loading additional dynamic libraries from input file

hugary1995:
- Unify registry data structure
- Add a Registry::load method to merge registry from dynamic libraries
- Add an option named "additional_libraries" under [Settings]


hugary1995:
@dschwen I am not sure if this sufficient for your use case. Could you review the `Registry::load` part?

reverendbedford (approved):

PR #243, opened 2025-04-02
New crystal plasticity formulation

hugary1995:
The new formulation uses multiplicative FeFp decomposition without approximation in kinematics.

This formulation should in theory be fully consistent with the MOOSE CP implementation.

- Added two new objects: R2Multiplication and GreenLagrangeStrain
- Added unit tests
- Added a regression test using the new formulation


tests/regression/pyrolysis/TGA/PR_pyrolysis.i
reverendbedford, line 1: Hey now, this shouldn't be here.

src/neml2/models/R2Multiplication.cxx
```diff
@@ -0,0 +1,91 @@
...
+
+namespace neml2
+{
+register_NEML2_object(R2Multiplication);
+
+OptionSet
+R2Multiplication::expected_options()
+{
+  OptionSet options = Model::expected_options();
+  options.doc() = "Multiplication of form \\f$ A B \\f$, where \\f$ A \\f$ and \\f$ B \\f$ are "
```
reverendbedford, line 35: Probably should note you can invert one or both in this docstring.

include/neml2/models/IsotropicSR2.h
```diff
@@ -0,0 +1,49 @@
...
+
+#pragma once
+
+#include "neml2/models/Model.h"
+
+namespace neml2
+{
+class SR2;
+
+class IsotropicSR2 : public Model
```
reverendbedford: Probably should have a short comment here explaining what it is.

reverendbedford (changes requested):
Couple of minor things.

python/tests/requirements.txt
```diff
@@ -1,3 +1,3 @@
-pyzag==1.0.0
+pyzag==1.1.1
```
hugary1995: Should this be `==` or a range?

python/examples/crystal_plasticity/compare_formulations.ipynb
hugary1995, line 1: > There are two common ways to formulate crystal plasticity kinematic:

kinematic -> kinematics

> Calculate the elastic deformation in postprocessing through the usual multiplicative decomposition of $F = F_e F_p$.

Maybe remove this sentence and say something like "both formulations are based on the multiplicative decomposition $F = F_e F_p$".

> separated and multplicative models

I think we should first define the namings. Something like "in this notebook we refer to the first formulation as 'separated' and the second as 'multiplicative'". I feel like these names are debatable though, because both a multiplicative one could argue.

hugary1995, line 1: I suggest using red dashed (as opposed to solid) lines in the plots. Adding legends would be nice.

tests/regression/pyrolysis/TGA/PR_pyrolysis.i
hugary1995: These are just formatting changes.

python/examples/crystal_plasticity/compare_formulations.ipynb
reverendbedford: I made some changes to the introduction and went with "separated" and "integrated".

reverendbedford: Yep.

reverendbedford (approved):

PR #244, opened 2025-04-11
Add ODF pole figure plotting

reverendbedford:
- Add ODF pole figure plotting routines: both the density calculation and the actual pretty plotting functions.
- Fix how we're doing ODF reconstruction to use papers published after the early 80s
- Miscellaneous math operators needed for the pole figure density calculation.


reverendbedford:
Let me add tests for at least the math stuff...  I'll try to do it this evening.

python/neml2/postprocessing/polefigure.py
```diff
@@ -90,6 +154,162 @@ def symmetry_operators_as_R2(orbifold, device=torch.device("cpu")):
...
+            vals.detach(),
+            cmap="Greys",
+            levels=torch.linspace(limits[0], limits[1], ncontour),
+            extend="both",
+        )
+    else:
+        CS = ax.contourf(mesh[0], mesh[1], vals.detach(), cmap="Greys")
+
+    # Make the graph nice
+    plt.ylim([0, projection.limit])
```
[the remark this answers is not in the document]
reverendbedford: You are a liar.

include/neml2/tensors/Rot.h
```diff
@@ -69,6 +69,17 @@ class Rot : public VecBase<Rot>
   /// Fill from standard Rodrigues parameters
   static Rot fill_rodrigues(const Scalar & rx, const Scalar & ry, const Scalar & rz);
 
+  /// Fill by forming the rotation between two vectors
+  static Rot fill_rotate_between(const Vec & v1, const Vec & v2);
```
hugary1995: I think `fill_` is redundant since this is a static method. So in practice
```cpp
auto r = Rot::fill_rotate_between(...);
```
doesn't contain more information than
```cpp
auto r = Rot::rotation_between(...);
```
BTW, I suppose you want `rotation_between` instead of `rotate_between`.
Also, "between" sounds purposely vague -- is it rotating from v1->v2 or v2->v1? How about
```cpp
static Rot from_to(const Vec & v1, const Vec & v2);
```
You could even add a suffix to distinguish between active and passive rotations.

Finally, I'd like all named constructors (static methods like this one) to have the `[[ nodiscard ]]` attribute.

These comments also apply to several other methods that were added before this PR. We can deal with those later.

include/neml2/tensors/Rot.h
```diff
@@ -69,6 +69,17 @@ class Rot : public VecBase<Rot>
...
+
+  /// Fill with axis/angle pairs
+  static Rot fill_axis_angle(const Vec & n, const Scalar & theta);
```
hugary1995: Similar comments as above. How about renaming this to
```cpp
static Rot from_axis_angle(const Vec & n, const Scalar & theta);
```

include/neml2/tensors/Rot.h
```diff
@@ -69,6 +69,17 @@ class Rot : public VecBase<Rot>
...
+
+  /// Fill with axis/angle pairs as a standard Rodrigues parameter, then convert to MRP
+  // This has a specialized use in integrating through ODFs, helping me avoid the need to calculate
+  // the arc length scaling term for our space
+  static Rot fill_axis_angle_standard(const Vec & n, const Scalar & theta);
```
hugary1995: Similar comments.

src/neml2/tensors/Rot.cxx
```diff
@@ -152,6 +153,36 @@ Rot::fill_random(unsigned int n)
...
+
+  return nn * t;
+}
+
+Rot
+Rot::fill_axis_angle_standard(const Vec & n, const Scalar & theta)
+{
+  auto nn = n / n.norm();
+  auto t = neml2::tan(theta / 2.0);
+  auto vn = nn * t;
```
hugary1995:
```suggestion
  auto vn = fill_axis_angle(n, theta * 2);
```

python/neml2/postprocessing/polefigure.py
```diff
@@ -32,10 +32,39 @@
 
 import matplotlib.pyplot as plt
 
+import scipy.spatial as ss
+
 from neml2 import crystallography
 from neml2 import tensors
 
 
+def arbitrary_hemispherical_quadrature(points):
```
hugary1995: We need some kwargs for abs and rel tolerance for "on equator".

hugary1995 (changes requested):

hugary1995 (approved):

PR #245, opened 2025-04-18
Add a basic language server

hugary1995:
Language server currently supports
- Formatting
- Hover over text

I have yet to decide the language server versioning scheme:
- use same version/hash as the neml2 library
- use same version as the vscode extension
- a standalone schema-based versioning


hugary1995:
I'll re-implement this now that we got rid of wasp.

PR #254, opened 2025-05-24
Miscellaneous improvements

hugary1995:
- ~`Tensor::operator=(torch::Tensor)` assumes the original (LHS) batch shape~
- ~`Tensor::operator=(neml2::TensorBase)` makes no assumption~
- ~`PrimitiveTensor::operator=(torch::Tensor)` infers batch shape~
- ~`PrimitiveTensor::operator=(neml2::TensorBase)` makes no assumption~

~Basic error checks are added to make sure shapes are compatible.~

~Note `PrimitiveTensor::operator=(torch::Tensor)` is disabled during tracing, because inferring batch shape makes it difficult to optimize the traced function graph.~

* More forward declaration
* Only include necessary header files

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -27,7 +27,8 @@
         "CMAKE_BUILD_TYPE": "Debug",
         "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
         "NEML2_PCH": "OFF",
-        "NEML2_TESTS": "OFF"
+        "NEML2_RUNNER": "ON",
+        "NEML2_PYBIND": "ON"
       }
     },
     {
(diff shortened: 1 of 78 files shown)


dschwen:
Besides the failing test this looks fine. Should we be worried about implicit casts here? Some assignment that is actually unintentional/wrong, but now compiles due to an intermediate implicit cast via torch:: Tensor?

hugary1995:
I think the four rules above cover all cases.

One potentially concerning case is assigning a primitive tensor to another primitive tensor of a different type.

For example, R2 = R4. It used to be a compile time error, but now becomes a run-time error, since checking base shape after assignment at run-time is the best I can do.

Maybe it's worth taking a step back: I'm adding these new assignment operators for SWIFT based on my understanding of what is currently breaking. Maybe we should first reconsider your use case, i.e., is adding assignment operator overload the best solution?

dschwen:
Look at my draft pr. I think that should fix my issue as well.

hugary1995:
k, as long as swift can be fixed that's all good. I repurposed this PR.

PR #264, opened 2025-06-27
Fixes to compile/run on aurora

reverendbedford:
Miscellaneous fixes required to get things to work on aurora.


CMakeLists.txt
```diff
@@ -63,6 +63,7 @@ set(NEML2_PYBIND OFF CACHE BOOL "Build NEML2 Python bindings")
 set(NEML2_DOC OFF CACHE BOOL "Build NEML2 documentation (html)")
 set(NEML2_WORK_DISPATCHER OFF CACHE BOOL "Enable NEML2 work dispatcher")
 set(NEML2_JSON OFF CACHE BOOL "Enable JSON support")
+set(AURORA_BUILD OFF CACHE BOOL "Build on ALCF Aurora machine")
```
hugary1995, line 66: I'd prefer a more general flag like `NEML2_IPEX`. I suppose there's nothing else specific to Aurora.

CMakeLists.txt
```diff
@@ -130,6 +131,13 @@ if(Torch_DOWNLOADED)
       install(DIRECTORY ${Torch_LINK_DIR} DESTINATION . COMPONENT libneml2)
 endif()
 
+# ----------------------------------------------------------------------------
+# Intel Extension for PyTorch
+# ----------------------------------------------------------------------------
+if (AURORA_BUILD)
+      find_library(IPEX_LIB intel-ext-pt-gpu PATHS ${INTEL_EXTENSION_FOR_PYTORCH_PATH}/lib NO_DEFAULT_PATH REQUIRED)
```
hugary1995, line 138:
```suggestion
      find_library(IPEX_LIB intel-ext-pt-gpu)
```
I suspect something like this will just work. You can then specify the path using `-DIPEX_LIB_ROOT=...`.

tests/dispatchers/CMakeLists.txt
```diff
@@ -1,5 +1,10 @@
 file(GLOB_RECURSE srcs CONFIGURE_DEPENDS *.cxx)
 add_executable(dispatcher_tests ${srcs})
-target_link_libraries(dispatcher_tests PRIVATE testutils)
+if (AURORA_BUILD)
+    find_package(Threads REQUIRED)
+    target_link_libraries(dispatcher_tests PRIVATE testutils Threads::Threads)
+else()
+    target_link_libraries(dispatcher_tests PRIVATE testutils)
+endif()
```
hugary1995, line 8: Why is this required?

src/neml2/drivers/ModelDriver.cxx
```diff
@@ -146,4 +177,12 @@ ModelDriver::diagnose() const
   Driver::diagnose();
   neml2::diagnose(*_model);
 }
+
+void
+ModelDriver::to(Device dev)
+{
+  _device = dev;
+  setup();
+}
```
hugary1995, line 186: Why do we need this method as well as `TransientRegression::to`?

hugary1995 (changes requested):

src/neml2/drivers/ModelDriver.cxx
hugary1995: Oh I added this... now I'm questioning myself. We need this to run regression tests on different devices.

hugary1995:
I feel like this is no longer needed. We can keep this branch and re-visit if necessary.

PR #266, opened 2025-07-02
Fix solidification issues and update advective stress

hdt5kt:
- Area under the SymmetricHermiteInterpolation is now one
- Regression test for solidification process
- Corrected advective stress for system with phase change


tests/regression/chemical_reactions/solidification/gold/check_heat_release.py
hugary1995, line 1: There's probably no need to check this script into the repository.

hugary1995 (changes requested):

hugary1995 (approved):

PR #272, opened 2025-07-09
Bundled model

hugary1995:
This PR introduces two major capabilities 
- object serialization 
- model bundling

The object serialization works by recording options consumed when creating an object and serializes the recorded options afterwards.

The model bundling capability is built upon that. A model can be bundled and stored on disk in a compressed format. The compressed file contains
- the serialized options that can be used to recreate the model
- a json configuration file describing the model metadata

The metadata can include high-level description of the bundled model, as well as user-facing descriptions of each variable and parameter. This addresses one of the user complaints on hard-to-interpret names like forces/S, state/internal/Ep, etc.

The bundled model can be deserialized using the BundledModel object which only needs the path to the compressed file. The BundledModel can be used like a usual neml2 model and can be composed with others as usual.


hugary1995:
I think this isn't the ideal solution. A better solution would be trace and store model as torch scripts, and provide a dedicated model to load such scripts. For a composed model, we will have graph breaks (at models that disable JIT). We need to handle graph breaks by splitting tracing into sub-graphs. When deserializing such bundled model, we need to re-create the graph breakers (ImplicitUpdate, looking at you).

PR #294, opened 2025-09-26
fix mask and add a test case with edge cases

zaheen1258:


include/neml2/models/LinearInterpolation.h
```diff
@@ -108,9 +109,22 @@ template <typename T2>
...
-  // Use advanced (boolean) indexing to select the correct interval
-  return T2(in.batch_expand_as(m).index({m})).batch_reshape(B);
+
+  auto expanded_in = in.batch_expand_as(m);
+
+  Scalar bool_mask = m;
+  if (bool_mask.data().dtype() != at::kBool)
+  {
+    bool_mask = Scalar(bool_mask.data().to(at::kBool), bool_mask.batch_sizes());
+  }
```
hugary1995, line 120: Do we ever get non-boolean `m`?

include/neml2/models/LinearInterpolation.h
```diff
@@ -108,9 +109,22 @@ template <typename T2>
...
+
+  auto result_data = expanded_in.index({bool_mask});
+
+  if (B.empty())
+    return T2(result_data.data().squeeze());
+  else
+    return T2(result_data).batch_reshape(B);
```
hugary1995, line 127: Can you briefly explain the logic behind this change?

hugary1995 (changes requested):

PR #296, opened 2025-10-02
add ScalarLinearInterpolation edge case unit tests

zaheen1258:
One is batched and the other is not. Both cases are worth keeping, as I've seen that some fixes would work for one and fail for the other.


hugary1995:
@codex review this PR.

hugary1995:
This should have been fixed #331. Feel free to reopen if otherwise.

PR #316, opened 2026-01-15
Refactor tensor derivatives and solver infrastructure; update python/pyzag bindings, tests, and docs

### LINKED ISSUE #116: Better way to deal with "lists" of tensors
We need to consider a third set of dimensions in our classes that are set at runtime.  As an example, a `BatchTensor` might have shape `(3,4) + (7,) + (6,)` where `(3,4)` is the batch dimension, `(7,)` is the "list" dimension, and `(6,)` is the base dimension.  So we could view this as a list of batched `SR4` tensors, for example.  By default all our current mathematical operators (except the list tensor stuff) will treat the "list" dimension as batch dimension.  But we can add new outer product type operators that apply binary operators to do list outer productions.  So if A is shape `(3,4) + (7,) + (6,)`  and B is shape `(3,4) + (5,) + (6,)`  `list_product(A, B, [](auto a, auto b){return a+b;})` would return a tensor with shape `(3,4) + (7,5) + (6,)` that adds the list entries of A and B together in all outer product combinations.

### LINKED ISSUE #117: Improve Taylor models and similar structures
We need a way to "wrap" a batch dimension in some reduction/expansion operations.  So a model of this type might expand the base model forces along some axis, run the batch, but only return the average result.  The external interface to this type of model is consistent -- you have the same batch shapes in and out.  The only real challenge is the "inner" model state, which will have one extra batch dimension and, for stateful models, need to persist from time step to time step.

Solving issue #116 may help with this.

hugary1995:
Summary: Major refactor of tensor derivative handling and equation‑system infrastructure, with solver/linear‑solver cleanups and broad updates across models, Python bindings, docs, and tests.

Changes:

- Replace model‑level assembler/derivative plumbing with new tensor derivative and equation‑system types
- Introduce linear solver abstraction and DenseLU implementation; remove trust‑region subproblem/assembler artifacts and related tests.

close #116 #117


include/neml2/equation_systems/HMatrix.h
reverendbedford, line 1: I think this needs some documentation.  It looks like this is always a dense block tensor right now?  That's fine, but maybe document the structure.

include/neml2/equation_systems/HVector.h
```diff
@@ -0,0 +1,116 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+#pragma once
+
+#include "neml2/equation_systems/HeterogeneousData.h"
+
+namespace neml2
+{
+std::vector<TensorShape> shape_refs_to_shapes(const std::vector<TensorShapeRef> & shape_refs);
```
reverendbedford: Why are these two necessary?

include/neml2/equation_systems/HVector.h
```diff
@@ -0,0 +1,116 @@
...
+std::vector<TensorShapeRef> shapes_to_shape_refs(const std::vector<TensorShape> & shapes);
+
+std::vector<std::size_t> select_subblock_indices(OptionalArrayRef<std::size_t> blocks,
```
reverendbedford: This needs some docs.  Is this doing "advanced indexing" to select blocks?

include/neml2/equation_systems/HVector.h
```diff
@@ -0,0 +1,116 @@
...
+                                                 std::size_t n);
+std::vector<TensorShapeRef> select_shapes(const std::vector<TensorShape> & all_shapes,
```
reverendbedford: Not sure what this does just from the header.

include/neml2/equation_systems/LinearSystem.h
```diff
@@ -0,0 +1,193 @@
...
+public:
+  LinearSystem() = default;
+  LinearSystem(const LinearSystem &) = default;
+  LinearSystem(LinearSystem &&) noexcept = default;
+  LinearSystem & operator=(const LinearSystem &) = default;
+  LinearSystem & operator=(LinearSystem &&) = default;
+  virtual ~LinearSystem() = default;
+
+  ///@{
+  /// Set the unknown u from the previous step
```
reverendbedford: What's g and gn?

include/neml2/equation_systems/LinearSystem.h
reverendbedford, line 1: I'll probably understand when I look at the implementation, but what are g and gn?  Probably needs a bit more class doc.

include/neml2/models/Model.h
```diff
@@ -39,12 +38,27 @@
...
+  AssemblyingNonlinearSystem(Model * const, bool assembling = true);
+
+  AssemblyingNonlinearSystem(const AssemblyingNonlinearSystem &) = delete;
+  AssemblyingNonlinearSystem(AssemblyingNonlinearSystem &&) = delete;
+  AssemblyingNonlinearSystem & operator=(const AssemblyingNonlinearSystem &) = delete;
+  AssemblyingNonlinearSystem & operator=(AssemblyingNonlinearSystem &&) = delete;
+  ~AssemblyingNonlinearSystem();
+
+  Model * const model;
+  const bool prev_bool;
```
reverendbedford, line 65: What does this do?

include/neml2/models/Reduction.h
reverendbedford, line 1: I noticed this in linear interpolation too, why do we have to reduce only along the last dim?

include/neml2/solvers/LinearSolver.h
reverendbedford, line 1: This is picky, but should it be `support_factorization` and `backsolve` or something?  If we have a Cholesky decomposition for example it would fit in this interface but it's not "LU".  Or SVD or whatever.

tests/regression/solid_mechanics/crystal_plasticity/taylor/model.i
reverendbedford, line 1: Does the "old taylor" not work anymore?

reverendbedford (changes requested):
Main complaint:

The linear system stuff is a confusing.  More docs will help somewhat (need to mention the names g = "givens", b = "RHS", "u" = unknowns.  I guess the last one is obvious.

Now that I read the source files I get what's happening, but isn't including givens in the linear system a bit of a mixed metaphor?  I guess we can discuss tomorrow because right now I can't think of a better way to structure things.

Random other questions:

1. Why did the trust region solver have to go?  We don't need it, just curious.
2. Why are reductions on intermediate dimension now always -1?

include/neml2/solvers/LinearSolver.h
hugary1995: Yeah but there would be factorizations that do not fit in this interface, e.g., LDL, etc. For now I removed these methods as they are not needed so far. When we need them, I'll add them back and discuss.

tests/regression/solid_mechanics/crystal_plasticity/taylor/model.i
hugary1995: I plan to add that back in a follow up PR

include/neml2/models/Reduction.h
hugary1995: For Reduction, we have to always deal with the last dimensions, because they are "intrinsic". For LinearInterpolation, we could technically support any dimension, because they do not appear in derivatives. IDK, if there's a strong need to support interpolating along arbitrary dimension, I guess I can add it back.

include/neml2/models/Model.h
hugary1995: This is the guard telling the model not to bother calculating derivatives w.r.t. forces (during assembly of the nonlinear system).

PR #318, opened 2026-01-21
Fix python binding memory leaks

hugary1995:
There are two leak sites:
1. In python/neml2/tensors/TensorBase.cxx, I was returning heap pointers hoping python would take ownership and delete them. But apparently that's not happening. I am switching to unique_ptr + keep_alive to make sure the views outlive the original tensor.
2. In python/neml2/tensors/types.h, the type casters make PySlice_New without stealing the ownership from unpacked integers. Switching to pybind11::slice to avoid leaks.

Thanks @Jes-s-L for reporting the bug!

### DIFF
--- a/python/neml2/csrc/tensors/BaseView.cxx
+++ b/python/neml2/csrc/tensors/BaseView.cxx
@@ -33,8 +33,7 @@ void
 def_BaseView(py::module_ & m, const std::string & name)
 {
   auto c = py::class_<BaseView<T>>(m, name.c_str());
-  c.def(py::init<T *>())
-      .def("dim", &BaseView<T>::dim)
+  c.def("dim", &BaseView<T>::dim)
       .def_property_readonly("shape",
                              [](const BaseView<T> & self)
                              {
@@ -72,8 +71,9 @@ def_BaseView(py::module_ & m, const std::string & name)
 }
 
 template <class T>
-BaseView<T>::BaseView(T * data)
-  : _data(data)
+BaseView<T>::BaseView(py::object data)
+  : _owner(std::move(data)),
+    _data(_owner.cast<T *>())
 {
 }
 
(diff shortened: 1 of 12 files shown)


hugary1995:
@Jes-s-L I only tested this on the toy problem you gave us. If you can confirm this patch on your real problem, that'd be great.

I had to make two changes in the input file to make it work with the newest version of neml2:
```python
[Tensors]
  [temp]
    type = Scalar
    values = '15. 25. 150. 250. 350. 450. 550. 600.'
    batch_shape = '(8)'
    intermediate_dimension = 1  # this line
  []
  [E_values]
    type = Scalar
    values = '212255 212115 206309 201665 195231 185384 172540 155328'
    batch_shape = '(8)'
    intermediate_dimension = 1  # this line
  []
[]
```
and
```python
  [mixed]
    type = MixedControlSetup
    above_variable = 'state/S'
    below_variable = 'state/E'
  []
  [mixed_old]
    type = MixedControlSetup
    control = 'old_forces/control'
    mixed_state = 'old_state/mixed_state'
    fixed_values = 'old_forces/fixed_values'
    above_variable = 'old_state/S'
    below_variable = 'old_state/E'
  []
```

Jes-s-L:
@hugary1995 this looks to have fixed it, the NEML2 model call no longer accumulates memory in my main script. Thank you!

reverendbedford:
@Jes-s-L try again please, but I think this works and shouldn't segfault.

Jes-s-L:
@reverendbedford Seems to work fine for me, but I just tested the pyzag/python side of things. Is there anything else I should test?

reverendbedford:
> Is there anything else I should test?

Hopefully our tests will cover everything else.  Thanks for pointing this out.

reverendbedford (approved):
I mean at this point I/Claude wrote it, I guess let's have the other AI weigh in.

PR #322, opened 2026-02-05
Updated dispatchers to remove TIMPI

reverendbedford:
Removes the TIMPI dependency and updates `SimpleMPIDispatcher` slightly.


src/neml2/dispatchers/SimpleMPIScheduler.cxx
```diff
@@ -113,18 +114,24 @@ SimpleMPIScheduler::determine_my_device()
   // NOLINTBEGIN(modernize-avoid-c-arrays)
   char c_str_hostname[MPI_MAX_PROCESSOR_NAME];
   int name_len = 0;
-  timpi_call_mpi(MPI_Get_processor_name(c_str_hostname, &name_len));
+  MPI_Get_processor_name(c_str_hostname, &name_len);
```
hugary1995: We probably should check for errors after calls like this. You can use a simple macro like
```cpp
#define neml2_call_mpi(call)                                                \
  do {                                                                      \
    int err = (call);                                                       \
    if (err != MPI_SUCCESS) {                                               \
      char err_string[MPI_MAX_ERROR_STRING];                                \
      int len = 0;                                                          \
      MPI_Error_string(err, err_string, &len);                              \
      throw std::runtime_error(std::string("MPI error: ") + err_string);    \
    }                                                                       \
  } while (0)
```

hugary1995: same thing applies to other MPI calls

CMakeLists.txt
```diff
@@ -372,6 +344,13 @@ if(NEML2_DOC)
       set(doxygen-awesome-css_SOURCE_DIR ${NEML2_CONTRIB_PREFIX}/doxygen-awesome-css-src)
 endif()
 
+# ----------------------------------------------------------------------------
+# MPI
+# ----------------------------------------------------------------------------
+if(NEML2_WORK_DISPATCHER)
+      find_package(MPI COMPONENTS CXX REQUIRED)
+endif()
+
```
hugary1995, line 353: Well this is not your fault, but I start to wonder if we can/should support dispatcher when MPI is absent. At least some dispatchers will still work.

hugary1995: I can work on that later.

hugary1995 (changes requested):

include/neml2/dispatchers/SimpleMPIScheduler.h
```diff
@@ -96,4 +96,17 @@ class SimpleMPIScheduler : public WorkScheduler
...
+#define neml2_call_mpi(call)                                                                       \
+  do                                                                                               \
+  {                                                                                                \
+    int err = (call);                                                                              \
+    if (err != MPI_SUCCESS)                                                                        \
+    {                                                                                              \
+      char err_string[MPI_MAX_ERROR_STRING];                                                       \
+      int len = 0;                                                                                 \
+      MPI_Error_string(err, err_string, &len);                                                     \
+      throw std::runtime_error(std::string("MPI error: ") + err_string);                           \
```
hugary1995: Sorry I gave you a bad snippet to begin with. I think we need to throw at least a NEMLException, as later I could be collecting stack trace at the throw site.

hugary1995 (approved):
One optional request, otherwise good to go.

include/neml2/dispatchers/SimpleMPIScheduler.h
reverendbedford: lol, fine.

PR #328, opened 2026-02-27
Need complete Derivative type for VariableBase destructor

lindsayad:
I don't know why this isn't triggering in other places but the compiler errored for me when building our MFEM NEML2 miniapp


hugary1995 (approved):
Seems like a typical "new" compiler issue (in a good way).

lindsayad:
It was while passing -std=c++20

hugary1995:
Interesting.

PR #336, opened 2026-04-08
Fix inverse pole figure projection

reverendbedford:
Fixes inverse pole figure projection so it actually works properly.  The main error is that unlike standard pole figures you need to include inversion in the symmetry operators.  Also restores Vec::dot which we used in the original routines.


python/neml2/postprocessing/polefigure.py
```diff
@@ -445,7 +449,20 @@ def __init__(
...
+        if n01.dot(v2).torch() < 0:
+            n01 = -n01
+
+        n12 = v1.cross(v2)
+        if n12.dot(v0).torch() < 0:
+            n12 = -n12
+
+        n20 = v2.cross(v0)
+        if n20.dot(v1).torch() < 0:
+            n20 = -n20
```
hugary1995, line 466: I didn't know we can convert a torch.Tensor to a boolean. I think for example `n01.dot(v2).torch < 0` gives you a boolean torch.Tensor. What happens if its batches have disagreeing booleans?

python/examples/crystal_plasticity/polefigures.ipynb
hugary1995, line 1: I think we are supposed to install the notebook pairing pre-commit hook. It will take care of syncing the notebooks to their corresponding markdowns. I should try to enforce that somehow...

I marked .ipynb files as binaries, so I'm also surprised that github renders its diff. I need to figure out what's going wrong.

hugary1995 (changes requested):

python/examples/crystal_plasticity/polefigures.ipynb
reverendbedford: Is this something I was supposed to do or something you need to look at?

python/neml2/postprocessing/polefigure.py
reverendbedford: It works fine for dim = 1 tensors, which is probably what we want here.  I added a check to ensure we don't violate that assumption.

hugary1995 (approved):

PR #394, opened 2026-06-25
CI: keep PyPI as a torch-dep fallback; guard/skip reverse-AD AOTI on torch 2.11

hugary1995:
Two unrelated CI breakages surfaced on the push to `main` after #392 merged (the PR itself was green because neither path runs/blocks on PRs).

## 1. Intermittent `ResolutionImpossible` when installing torch

Doc builds and some compat jobs failed with e.g. `No matching distribution found for mpmath` / `fsspec`. Root cause: the torch install used

```
pip install torch==X --index-url https://download.pytorch.org/whl/cpu
```

`--index-url` makes the PyTorch CPU index the **only** source, so torch's transitive deps that the index doesn't mirror (`mpmath`, `fsspec`, …) fail to resolve whenever the index lags — purely intermittent, which is why it never showed on the PR.

**Fix:** add `--extra-index-url https://pypi.org/simple` so the `+cpu` torch wheel still comes from the PyTorch index (higher local-version precedence) while its deps fall back to PyPI. Applied in:
- `.github/actions/neml2-ci/action.yaml` (the Linux CPU torch install)
- `.github/workflows/compat.yaml` (the runtime-torch install)

## 2. torch 2.11.0 systematically fails the AD/AOTI tests

All 10 torch 2.11.0 compat jobs fail the parameter-derivative / `request_AD` AOTI tests with `torch._dynamo.exc.Unsupported: Unsupported Tensor.requires_grad_() call` — a torch-2.11.0 dynamo-lowering regression (2.10.0 and 2.12.x are fine). This is masked on PRs because the matrix only runs 2.11.0 on push/schedule.

**Fix:** drop the 2.11.0 rows from `doc/content/installation/compatibility.yaml` and re-render the doc table. The advertised range stays `[2.10.0, 2.12.1]`; 2.11.0 is the unsupported point in between (documented inline).

## Verification
- `python scripts/compat_matrix.py check` → OK (30 rows, range `[2.10.0, 2.12.1]`).
- `python scripts/dep_manager.py check` → OK.
- pre-commit (GH Actions lint, dep consistency, compat render/check) passes.

### DIFF
--- a/.github/actions/neml2-ci/action.yaml
+++ b/.github/actions/neml2-ci/action.yaml
@@ -52,7 +52,10 @@ runs:
       if: ${{ runner.os == 'Linux' && inputs.cpu-torch == 'true' }}
       shell: bash
       run: |
-        pip install torch==${{ inputs.torch-version }} --index-url https://download.pytorch.org/whl/cpu
+        # --extra-index-url keeps PyPI as a fallback: --index-url alone makes the
+        # PyTorch CPU index the *only* source, so torch deps it doesn't mirror
+        # (mpmath, fsspec, ...) intermittently fail to resolve.
+        pip install torch==${{ inputs.torch-version }} --index-url https://download.pytorch.org/whl/cpu --extra-index-url https://pypi.org/simple
 
     - name: Install PyTorch + nvcc + CCCL headers (Linux, with CUDA)
       # Inductor's CUDA JIT / AOT codegen needs three things on top of
--- a/.github/workflows/compat.yaml
+++ b/.github/workflows/compat.yaml
@@ -123,7 +123,9 @@ jobs:
         # torch version constraint, so the pinned version sticks.
         run: |
           if [ "${{ runner.os }}" = "Linux" ]; then
-            pip install torch==${{ matrix.torch }} --index-url https://download.pytorch.org/whl/cpu
+            # --extra-index-url keeps PyPI as a fallback so torch deps the PyTorch
+            # CPU index doesn't mirror (mpmath, fsspec, ...) still resolve.
+            pip install torch==${{ matrix.torch }} --index-url https://download.pytorch.org/whl/cpu --extra-index-url https://pypi.org/simple
           else
             pip install torch==${{ matrix.torch }}
           fi
--- a/neml2/cli/aoti_export.py
+++ b/neml2/cli/aoti_export.py
@@ -1758,6 +1758,41 @@ _AOTI_PARAM_DERIV_BUG_HINT = (
 )
 
 
+def _reverse_ad_aoti_unsupported_reason() -> str | None:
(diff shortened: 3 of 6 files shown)


hugary1995:
### Why torch 2.11.0 is guarded/skipped here — the upstream story

The 2.11.0 failures (`torch._dynamo.exc.Unsupported: Unsupported Tensor.requires_grad_() call`) come from our reverse-mode-AD AOTI path, which compiles under `torch._dynamo.config.trace_autograd_ops` and traces `raw.clone().requires_grad_(True)` before consuming `torch.autograd.grad` *inside* the segment. Whether dynamo can trace that `requires_grad_()` call has flip-flopped across torch releases:

- **Banned (~torch 2.1):** pytorch/pytorch#110053 "[dynamo] Add graph break on requires_grad_()" made `requires_grad_()` a hard graph break. It fixes pytorch/pytorch#107861, where dynamo *silently dropped* `requires_grad` on a `clone().requires_grad_()` intermediate (a variant of pytorch/pytorch#90552): calling it on a graph intermediate turns the tensor into an autograd leaf, and AOTAutograd's functionalization then drops the effect → silently wrong results. Graph-breaking was the safe stopgap (bdhirsh/eellison in the thread).
- **Re-enabled (torch 2.12, merged 2026-03-18):** pytorch/pytorch#176984 "[dynamo] Support requires_grad_ and then backward call" traces `requires_grad_()` again *when the autograd effect is consumed within the compiled region*, adding a correctness guard (`output_graph._check_requires_grad_intermediate_outputs`) that errors only if such an intermediate *leaks as a graph output* — catching the #107861 failure loudly instead of silently.
- **2.11.0 is the gap:** `trace_autograd_ops` already exists there, but #176984 hadn't landed yet, so our path hits the old graph break. (2.10 lacks `trace_autograd_ops` entirely — those tests skip; 2.12 has the fix — they pass.)

**Implications for neml2**
- The skip/guard here reflects an upstream gap, not a neml2 bug. `_reverse_ad_aoti_unsupported_reason()` is shared by the exporter's runtime guard and the test skips so the two can't drift.
- We stay on the safe side of the 2.12 guard: `autograd.grad` is consumed inside the segment and only plain (detached) Jacobian values are returned — so we must never let a `requires_grad_()` intermediate escape the exported graph as an output.
- The broader autograd-in-dynamo capture is still maturing upstream (open: pytorch/pytorch#185034 capture `autograd.grad`, pytorch/pytorch#185033 `Tensor.backward` by default, pytorch/pytorch#184958 escaped vjp wrappers), so `trace_autograd_ops` stays opt-in for now and version-guarding remains the right tool.

Refs: pytorch/pytorch#176984 (re-enable), pytorch/pytorch#110053 + pytorch/pytorch#107861 (original ban).

hugary1995:
code path is lean and hard to cover in unit test -- codecov check is allowed to fail

PR #395, opened 2026-06-25
Fix norm(SR2) eps regularizer to match v2 (unsquared)

hugary1995:
## Summary

`norm(A: SR2, eps)` was computing `sqrt(norm_sq(A) + eps*eps)`, **squaring** the regularizer under the sqrt. The v2 C++ `neml2::norm` (`src/neml2/tensors/functions/norm.cxx`) adds `eps` **unsquared**:

```cpp
Scalar norm(const Tensor & a, const std::optional<CScalar> & eps)
{
  return eps ? sqrt(norm_sq(a) + eps.value()) : sqrt(norm_sq(a));
}
```

The squared form claimed v2 parity in its comment but diverged from it. This PR uses plain `eps` under the sqrt to restore v2 behavior.

## Impact

The only callers (`AssociativeJ2FlowDirection`, `PowerLawKinematicHardeningStaticRecovery`, `ChabochePlasticHardening`) pass `eps = torch.finfo(dtype).eps` (machine precision). Input files never set this regularizer — it is sourced internally to match the C++ `machine_precision(dtype)` regularization. The change is therefore confined to the at-zero regularization value (e.g. for float64, `norm(0, eps)` goes from `eps ≈ 2.2e-16` to `sqrt(eps) ≈ 1.5e-8`); away from zero, both addends are far below float64 round-off relative to `norm_sq`.

## Changes

- `neml2/types/functions.py` — `eps * eps` → `eps` in the SR2 branch of `norm`; docstring now states `sqrt(norm_sq(A) + eps)` explicitly.
- `tests/unit/test_types_sr2.py` — `test_norm_eps_protects_zero` now pins `sqrt(eps)` (was the squared `eps`).

## Test plan

- [x] `tests/unit/test_types_sr2.py`, `tests/unit/test_normality.py`
- [x] full `tests/unit` + `tests/models` sweep (running locally)
- [x] `tests/regression` + `tests/verification` sweep (running locally)

### DIFF
--- a/neml2/types/functions.py
+++ b/neml2/types/functions.py
@@ -346,16 +346,17 @@ def norm(A, eps: float = 0.0):
       The ``eps`` regularizer (added unsquared under the sqrt) keeps the
       result differentiable at ``A == 0``; matches ``neml2::norm`` on
       the v2 C++ side.
-    * ``norm(t.base)`` -- ``sqrt(sum_over_base(t * t))`` on a
-      :class:`~neml2.types.Tensor`. Returns a ``Tensor`` with
-      ``base_ndim=0``; ``batch`` / ``sub_batch`` axes are preserved.
+    * ``norm(t.base, eps=0.0)`` -- ``sqrt(sum_over_base(t * t) + eps)`` on
+      a :class:`~neml2.types.Tensor`. The same unsquared ``eps`` regularizer
+      applies. Returns a ``Tensor`` with ``base_ndim=0``; ``batch`` /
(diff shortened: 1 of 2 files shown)


hugary1995:
Local sweeps all green:
- `tests/unit` + `tests/models`: **710 passed**
- `tests/regression` + `tests/verification`: **81 passed**

No gold values shifted — confirms the change is confined to the at-zero regularization value (callers pass `eps = finfo(dtype).eps`).

PR #398, opened 2026-06-26
Fix AOTI binding failing to import under cpu-only torch

hugary1995:
## Problem

On a **cpu-only torch** (e.g. Colab's `torch==2.11.0+cpu`), loading a `neml2-compile` stub via `neml2.load_model` raised:

```
KeyError: "Type 'AOTIModel' is not registered in NativeRegistry ..."
```

That `KeyError` is a downstream symptom. The actual failure (from the Colab traceback) is:

```
ImportError: libc10_cuda.so: cannot open shared object file: No such file or directory
```

## Root cause

The **published wheel** hard-linked `torch::cuda`, so `libneml2.so` / `_aoti.so` carried a `DT_NEEDED` on `libc10_cuda.so`. The link came from `CMakeLists.txt` (`if(TARGET torch::cuda) target_link_libraries(... torch::cuda)`), which fires because cibuildwheel's `before-build` installs `torch` from the **default PyPI index = the CUDA build**. A cpu-only torch ships no `libc10_cuda.so`, so the binding can't `dlopen`; `neml2/__init__.py`'s `except ImportError: pass` then swallows it and `AOTIModel` is never registered.

The cuda link was **spurious**: `nm -D --undefined-only libneml2` shows zero cuda symbols and `grep -r 'c10::cuda|at::cuda|CUDAStream' neml2/csrc/` is empty. neml2 reaches the GPU only through torch's device-generic dispatch (torch loads `libtorch_cuda` itself at runtime), so a single cuda-free `libneml2` runs under both cpu and cuda torch. Verified locally: a `--device cuda` artifact runs on `cuda:0` through the cuda-free library.

### Why CI missed it

Every lane was single-channel: `compat` builds + tests on the cpu index (never links cuda); cibuildwheel builds + self-tests on the default-index cuda torch (so `libc10_cuda` is present in its own test env → green). Nobody imported a **cuda-built wheel under cpu torch** — the Colab configuration.

## Fix

- **`CMakeLists.txt`** — drop the `torch::cuda` link on both the `aoti` and `eager` targets (kept `OPTIONAL_COMPONENTS cuda` for future use).
- **`pyproject.toml`** — build + test cibuildwheel against the cpu torch index on Linux (smaller/faster, cuda-free by construction).
- **`scripts/compat_matrix.py` + `.github/workflows/compat.yaml`** — add a runtime `cpu|cuda` channel to the test matrix. **Full mode** (push to main / dispatch) additionally imports + cpu-tests the wheel under a **cuda** torch (Linux torch-extremes @ newest Python); PRs skip it (the cuda wheel is ~2.5 GB to install). Harden the import check to `from neml2.aoti import Model`, since plain `import neml2` swallows a binding-load failure. GPU kernels are not exercised (GitHub runners have no GPU).
- **`neml2/__init__.py`** — when the AOTI binding is present on disk but fails to import, emit a warning with the underlying exception instead of silently swallowing it.

## Test plan

- [x] Local rebuild: `libneml2` `DT_NEEDED` reduced to `libc10` + `libtorch_cpu`; `from neml2.aoti import Model` imports; `--device cuda` artifact runs on `cuda:0`.
- [x] `compat_matrix.py expand`: PR mode adds 0 cuda rows; full mode adds 2; `check` / `render` unaffected.
- [x] `dep_manager check`, `ruff`, actionlint (pre-commit) all pass.
- [ ] CI: the new full-mode cuda rows confirm the cpu-built wheel imports + cpu-tests under cuda torch.

### DIFF
--- a/.github/workflows/compat.yaml
+++ b/.github/workflows/compat.yaml
@@ -122,7 +122,14 @@ jobs:
         # pip's resolver from upgrading torch on us: pyproject.toml has no
         # torch version constraint, so the pinned version sticks.
         run: |
-          if [ "${{ runner.os }}" = "Linux" ]; then
+          if [ "${{ matrix.channel }}" = "cuda" ]; then
+            # cuda-runtime row (Linux only): install the default-index CUDA build.
+            # The wheel was built against CPU torch; this asserts it still imports
+            # and cpu-tests cleanly under a cuda torch -- the ABI / DT_NEEDED check
+            # that the cpu-only AOTI-import bug would have failed. GPU kernels are
+            # NOT exercised (GitHub runners have no GPU).
+            pip install torch==${{ matrix.torch }}
+          elif [ "${{ runner.os }}" = "Linux" ]; then
             # --extra-index-url keeps PyPI as a fallback so torch deps the PyTorch
             # CPU index doesn't mirror (mpmath, fsspec, ...) still resolve.
             pip install torch==${{ matrix.torch }} --index-url https://download.pytorch.org/whl/cpu --extra-index-url https://pypi.org/simple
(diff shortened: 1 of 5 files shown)


hugary1995:
code coverage is expected to be 0.

PR #419, opened 2026-07-08
Port pyzag adapter to neml2 v3

hdt5kt:
co-authored with Claude - Opus 4.8

this should work with pyzag 2.0.0 (newest released)

Rework the pyzag integration for the neml2 v3 API and pyzag 2.0.0:

- neml2/pyzag/interface.py: replace NEML2PyzagModel with NEML2PyzagFactory, a pyzag NonlinearFunctionOperatorFactory over a ModelNonlinearSystem.
- neml2/pyzag/operators/: new pyzag block-operator backend backed by neml2 AssembledMatrix/AssembledVector -- BlockVector, SolvableBlockOperator (cached batched-LU solve + single-group-dense PCR), BlockJacobian, CachingLU, and typed assembly/transpose/select helpers; unsupported sub-batch (>=1 intermediate-axis coupling) layouts are rejected with a clear error.
- neml2/types/_boundary.py: add the to_torch() typed->raw pyzag boundary helper.
- neml2/__init__.py: expose the adapter surface.
- deps: bump pyzag 1.1.3 -> 2.0.0 (pyproject.toml, scripts/dependencies.yaml, .github/workflows/compat.yaml).
- tests: rename test_NEML2PyzagModel.py -> test_neml2_pyzag_factory.py; add conftest.py, test_adjoint_multigroup.py, test_neml2_backend.py; update the adjoint/compile tests.
- examples/docs: move the pyzag examples under neml2/pyzag/examples/; refresh the optimization tutorials and crystal-plasticity notebooks.
- added the lifted pcr for taylor approximation


neml2/types/_boundary.py
```diff
@@ -54,6 +54,17 @@
...
 
+def to_torch(t: TensorWrapper | Tensor) -> torch.Tensor:
+    """Return the raw ``torch.Tensor`` backing a typed wrapper.
+
+    The pyzag adapter (``neml2/pyzag/``) has to hand raw tensors to pyzag at its
+    framework boundary. Rule 2 forbids ``.data`` access outside ``neml2/types/``,
+    so the adapter calls this helper instead of reaching for ``.data`` itself --
+    the single sanctioned typed->raw read stays here, in the boundary module.
+    """
+    return t.data  # data-ok pyzag adapter boundary
```
hugary1995: **Remove this wrapper.** `to_torch()` is just `return t.data` with the marker relocated into `neml2/types/`. Every existing sanctioned boundary accesses `.data` *directly at the call site* with a `# data-ok <reason>` marker (see `es/implicit.py`, `models/param_ad.py`, `cli/aoti_export.py`) — the shared `_boundary.py` helpers are for reusable multi-step assembly, not for aliasing `.data`. Please drop this and switch call sites to direct `.data  # data-ok pyzag boundary` where a raw tensor genuinely crosses into pyzag, and to typed accessors everywhere else (see the metadata comment on `_operator.py`).

neml2/pyzag/operators/_operator.py
```diff
@@ -0,0 +1,428 @@
...
+
+    @property
+    def nblk(self) -> int:
+        """Number of blocks along the dynamic (time) axis."""
+        return to_torch(self.am.tensors[0][0]).shape[0]
+
+    @property
+    def batch_size(self) -> int:
+        """Size of the plain (non-dynamic) batch axis."""
+        return to_torch(self.am.tensors[0][0]).shape[1]
```
hugary1995: **Metadata reads should use the typed accessors, not `.data`.** These unwrap to raw purely to read `.device`/`.dtype`/`.shape` — no tensor crosses into pyzag here, so this isn't a boundary use. `Tensor` already exposes `.device`, `.dtype`, `.shape`, `.ndim`, `.batch_shape`; read them off the typed block directly (`self.am.tensors[0][0].device`, …). Same at line 238, and in `_jacobian.py:106-128` (`_first_diag_raw` and the properties built on it — that helper can then be deleted).

neml2/pyzag/operators/_jacobian.py
```diff
@@ -0,0 +1,218 @@
...
+            Tensor(
+                to_torch(t).flip(0),
+                batch_ndim=t.batch_ndim,
+                sub_batch_ndim=t.sub_batch_ndim,
+            )
+            for t in row
+        ]
+        for row in am.tensors
+    ]
+    return AssembledMatrix(am.row_layout, am.col_layout, blocks)
```
hugary1995, line 42: **Manipulating a typed object with raw ops + rewrap — use a typed primitive.** `_flip_time` takes an `AssembledMatrix` and returns one (typed on both ends), yet drops to `to_torch(t).flip(0)` + manual rewrap — exactly the raw-op-then-rewrap pattern Rule 2 guards against. The typed API has no `flip`/`clone`/`pad`/`contiguous` today, so per Rule 2 please add the missing primitive in `neml2/types/` (e.g. a `batch.flip`) and call it here. Same pattern in `_operator.py` `clone()` (290-305) and `pad_front()` (348-374).

neml2/pyzag/operators/_cache.py
```diff
@@ -0,0 +1,80 @@
...
+    def solve(self, A: AssembledMatrix, b):
+        """Solve ``A x = b`` for a single-group ``A`` (vector or matrix RHS)."""
+        if A.row_layout.ngroup != 1 or A.col_layout.ngroup != 1:
+            raise ValueError("CachingLU only supports single-group layouts")
+        lu, piv = self._factor(A.tensors[0][0])
+        if isinstance(b, AssembledVector):
+            return AssembledVector(A.col_layout, [self._lu_solve(lu, piv, b.tensors[0])])
+        return AssembledMatrix(
+            A.col_layout, b.col_layout, [[self._lu_solve(lu, piv, b.tensors[0][0])]]
+        )
```
hugary1995: **Duplicates `DenseLU`, and there are two LU caches in this PR; also an `id()`-keying hazard.** `CachingLU.solve` (71-80) is `DenseLU.solve` (`solvers/dense_lu.py:67-77`) plus a factorization cache, and `NEML2SolvableBlockOperator._ensure_lu` (`_operator.py:124-134`) carries a *second* LU cache. Please consolidate: add caching to `DenseLU` (or a shared helper in `neml2/solvers/`) and have both paths use it. Separately, `_factor` keys on `id(raw)` (line 55) — Python reuses `id()` after GC, so a freed block's id can be reassigned and return a **stale factorization**. Key on something stable, or invalidate explicitly.

neml2/pyzag/operators/_operator.py
```diff
@@ -0,0 +1,428 @@
...
+                accum = torch.zeros(
+                    (nblk, sbat, *intmd_sizes, n_out_var),
+                    dtype=x.raw_tensors[0].dtype,
+                    device=x.raw_tensors[0].device,
+                )
+            out_tensors[i] = accum
+
+        filled = cast(list[torch.Tensor], out_tensors)  # every group is set above
+        out_intmd_dims = [max(0, t.ndim - 3) for t in filled]
+        return NEML2BlockVector(filled, out_layout, out_intmd_dims)
```
hugary1995: **Consider lifting this into a typed ES op.** ~85 lines of hand-rolled block matvec on raw tensors, deliberately different from `AssembledMatrix.__matmul__` (keeps BLOCK output per-site instead of aggregating). The semantics are legit and well-documented, but "per-instance / grain-diagonal matmul" is a linear-algebra primitive — it belongs as a typed mode/method on `AssembledMatrix` in `neml2/es/`, where it's unit-testable and reusable. Reserve the raw hand-off for the final `NEML2BlockVector` construction.

neml2/pyzag/operators/_assembly.py
```diff
@@ -0,0 +1,80 @@
...
+    return AssembledMatrix(
+        am.row_layout,
+        am.col_layout,
+        [[t.batch[key] for t in row] for row in am.tensors],
+    )
+
+
+def _select_dynamic_av(av: AssembledVector, key) -> AssembledVector:
+    """Index or slice the leading dynamic (time) axis of every group tensor as a view."""
+    return AssembledVector(av.layout, [t.batch[key] for t in av.tensors])
```
hugary1995: **These belong in the ES layer, not the adapter.** `_transpose_am`, `_select_dynamic_am`, `_select_dynamic_av` are generic `AssembledMatrix`/`AssembledVector` operations (transpose, batch/dynamic indexing), already written in typed terms, and `AssembledMatrix` currently has neither transpose nor indexing. Please promote them to `AssembledMatrix.transpose()` / `AssembledMatrix.batch[...]` / `AssembledVector.batch[...]` in `neml2/es/assembled.py`. (`_require_le_one_intmd` at 41-51 can stay as a backend guard, or move next to `AssembledMatrix`.)

neml2/pyzag/operators/_flat.py
```diff
@@ -0,0 +1,91 @@
...
+    for g, gflat in enumerate(group_flats):
+        if layout.structure[g] == "block":
+            sub_shape = tuple(int(s) for s in layout.group_sub_batch_shape(g))
+            sub_numel = prod(sub_shape) or 1
+            base = gflat.shape[-1] // sub_numel
+            reshaped = gflat.reshape(*gflat.shape[:-1], *sub_shape, base)
+            tensors.append(Tensor(reshaped, batch_ndim=dyn_dim, sub_batch_ndim=len(sub_shape)))
+        else:
+            tensors.append(Tensor(gflat, batch_ndim=dyn_dim, sub_batch_ndim=0))
+    return AssembledVector(layout, tensors)
```
hugary1995: **Layout math belongs on `AxisLayout`.** `_group_intmd_dim`, `_group_flat_size`, `_layout_flat_size` are pure functions of `AxisLayout` (DOF counts) — please move them onto `AxisLayout` in `neml2/es/axis_layout.py`. The flat pack/unpack (`_av_to_flat`/`_split_flat_to_av`) is more defensibly a pyzag-boundary concern, but would be cleaner and testable as `AssembledVector.to_flat()` / `AssembledVector.from_flat(layout)` on the ES type.

neml2/pyzag/interface.py
```diff
@@ -22,110 +22,98 @@
...
+            raw = to_torch(t)
+            cur = t.batch_ndim
+            while cur < target_ndim:
+                raw = raw.unsqueeze(0)
+                cur += 1
+            full = tuple(target_dynamic_shape) + tuple(raw.shape[target_ndim:])
+            raw = raw.expand(full).contiguous()
+            new_row.append(Tensor(raw, batch_ndim=target_ndim, sub_batch_ndim=t.sub_batch_ndim))
+        blocks.append(new_row)
+    return AssembledMatrix(am.row_layout, am.col_layout, blocks)
```
hugary1995, line 95: **Per-iteration full copy of the diagonal Jacobian.** `.expand(full).contiguous()` (line 98) materializes the entire diagonal Jacobian (`nstep × batch × …`) into a fresh contiguous tensor on every `evaluate_raw`, i.e. every Newton iteration. Please confirm the `.contiguous()` is actually required downstream; if only some blocks are non-contiguous, copy just those. (Related: `__setitem__` at `_operator.py:324-346` clones a whole block to write one slice — same copy-on-write concern.)

neml2/pyzag/examples/calibration/neml2_calibration.py
```diff
@@ -0,0 +1,331 @@
...
+def random_orientations(ngrains, dtype, device):
+    """Random uniform SO(3) grain orientations as ``(ngrains, 3)`` MRPs."""
+    g = torch.randn(ngrains, 3, 3, dtype=dtype, device=device)
+    q, r = torch.linalg.qr(g)
+    sign = torch.sign(torch.diagonal(r, dim1=-2, dim2=-1))
+    q = q * sign.unsqueeze(-2)
+    det = torch.linalg.det(q)
+    q[:, :, 0] = q[:, :, 0] * det.unsqueeze(-1)
+    r2 = neml2.types.R2(q, 1)
+    return neml2.types.MRP.from_matrix(r2).data.to(dtype=dtype, device=device).contiguous()
```
hugary1995: **User/tutorial code must not touch `.data`.** `neml2.types.MRP.from_matrix(r2).data...` — the `.data` boundary marker is for framework-boundary code, never for user-facing examples/notebooks (and `to_torch`/`Tensor.torch()` would be the same violation with a hat on). Use the public typed API; if an example genuinely needs a raw tensor and the public surface can't provide one, that's an API gap to fix, not to paper over here. Resolves naturally as part of moving these examples out of `neml2/`.

neml2/pyzag/interface.py
```diff
@@ -134,8 +122,13 @@ def __init__(
         *args,
         exclude_parameters: list[str] | None = None,
         include_parameters: list[str] | None = None,
+        compile: bool = True,
```
hugary1995: **Please don't compile in the constructor — keep `neml2.compile` the explicit opt-in.** `compile=True` by default is a new behavior (the class on `main` never compiled), and `neml2.compile` is already the explicit public verb for this (it duck-types the factory via `.sys`). Compiling in `__init__` creates a second way to do it, couples construction to an expensive/failure-prone `torch.compile` (needs `dynamic=False` + `allow_autograd`, can raise), and makes constructing the object for a test/introspection pay that cost — the `compile=False` "oracle" escape hatch is itself the tell. Please remove the `compile` parameter entirely: construction yields a py-eager model, users opt into py-jit via `neml2.compile(model)`. Move the dynamo cache-limit bump (currently in `_compile()`, 160-173) into `neml2.compile`'s pyzag branch — `compile.py` doesn't set those limits today.

hugary1995 (commented):
Thanks for this — the port is substantial and the lifted-arrowhead PCR is genuinely nice work. Most of the feedback below is about keeping raw-tensor use confined to pyzag's contract and pushing generic linear-algebra/layout helpers down into the `es`/`types` layers where they can be tested and reused. Inline comments carry the specifics; the cross-cutting items are here.

**Guiding principle for the `.data` items (1–8):** `.data` is sanctioned *only* where a raw tensor actually crosses into pyzag's ABC surface (constructing/consuming `BlockVector.raw_tensors`, feeding `torch.linalg`, returning to a pyzag ABC method). It is **not** sanctioned for reading metadata or for manipulating neml2-typed objects — those must use the typed API, adding a primitive if one is missing.

---

**General — replace `to_torch()` with direct `.data # data-ok pyzag boundary`, only at real boundary points.** After removing the wrapper (inline comment 1), use `x.data  # data-ok pyzag boundary` *only* at genuine contract points (e.g. `_vector.py:93`, `_flat.py:72`, `interface.py:479`). This matches how every existing boundary in the tree already does it (`es/implicit.py`, `models/param_ad.py`, `cli/aoti_export.py`). Note `_operator.py:421,427` read `.data` on pyzag's own `DenseBlockOperator`/`DenseBlockVector` — those are pyzag types, not neml2 wrappers, so they're fine as-is and need no marker.

**General — move `neml2/pyzag/examples/` out of the package; they ship in the wheel.** The 6 example files (+1,440 lines) live under `neml2/`, and `wheel.packages = ["neml2"]` (`pyproject.toml:27`) with the only exclude being `neml2/csrc/**` — so these standalone scripts ship inside every user's installed wheel. They're not a package (no `__init__.py`), not imported, and unreferenced anywhere. Please relocate: `.i` files a test needs → `tests/unit/pyzag/models/`; the narrative/plot drivers → `doc/content/tutorials/` (the optimization tutorials already updated in this PR) or a top-level `examples/` excluded from the wheel. Nothing demonstrational should live under `neml2/`. (Related: user/tutorial code must never touch `.data` — see inline comment on the calibration example.)

**General — please keep the public name `NEML2PyzagModel`.** Understood the class now implements pyzag 2.0's `NonlinearFunctionOperatorFactory` (vs. the old `NonlinearRecursiveFunction`), which motivated `…Factory`. For user-facing continuity we'd like to retain `NEML2PyzagModel`. Please revert the rename across `interface.py:104`, `neml2/pyzag/__init__.py`, `neml2/models/compile.py`, `doc/content/deployment/overview.md`, and the docstrings — the base class can change while the exported name stays. Note in the docstring that it implements the factory ABC.

**General — enumerate pyzag as a sanctioned boundary in CLAUDE.md.** To keep the hard rules and the code in agreement, add the pyzag adapter to the Rule 1 list and the Rule 2 exception paragraph, scoped to *fulfilling pyzag's ABC contract only* (metadata reads and typed-object manipulation stay typed). This can land in the same PR as the `pyzag/` layout-bullet edit you already made.

**Scope note:** inline comments 4, 6, 7, 8 ask for new code in `neml2/es/` and `neml2/types/` (typed primitives + ES methods). That's the right home, but it widens the PR — happy for those to be a fast follow-up with the pyzag backend switching to the new primitives once they land, if you'd prefer to keep this PR focused. Your call.

doc/content/deployment/overview.md
```diff
@@ -144,8 +144,10 @@ on one of the runtimes above:
   `ModelUnitTest`, `TransientRegression`, `Verification`) step a model through a
   load history on `py-eager`. If the input file names an `AOTIModel`, the same
   driver runs on `py-aoti` instead.
-- The **pyzag** adapter (`NEML2PyzagModel`) drives calibration on `py-eager`,
-  optionally accelerated to `py-jit` via `neml2.compile`.
+- The **pyzag** adapter (`NEML2PyzagFactory`) drives calibration on `py-jit` by
+  default -- it compiles the residual model in place via `neml2.compile` on
+  construction -- and falls back to `py-eager` when constructed with
+  `compile=False` (a correctness oracle).
```
hugary1995: **Revert this once compile-on-construction is removed.** Tied to the constructor comment — this should go back to "drives calibration on **py-eager**, optionally accelerated to py-jit via `neml2.compile`," since the constructor will no longer compile by default.

neml2/pyzag/operators/_cache.py
hdt5kt: - Shared helper: added LUCache in neml2/solvers/_lu_cache.py (exported from neml2.solvers). It factors once and memoizes (LU, pivots).
- id() fix: it keys on the source tensor held by reference (identity) plus its _version, with an explicit invalidate(). Holding the reference is what closes the hole — a freed block's id can no longer be reassigned to a stale entry, and _version catches in-place edits.
- Dedup: extracted DenseLU._solve_block (behavior-identical); CachingLU now subclasses DenseLU and overrides only that, using LUCache — the copied solve() and the hand-rolled id-keyed cache are gone.
- Both paths: the Schur primary solver (CachingLU) and the operator's Thomas path (_ensure_lu, injecting the CUDA-guarded factor) now share LUCache; __setitem__ invalidates it.

The DenseLU change is a pure extract of _solve_block (no behavior change, covered by tests/unit/test_solvers.py). Updated test_caching_lu_reuses_factorization to assert reuse via the new cache (identity of _src/_lu) instead of _key.

neml2/pyzag/operators/_operator.py
hdt5kt: lifted it into a typed ES op with added tests

neml2/pyzag/operators/_assembly.py
hdt5kt: - Promoted to neml2/es/assembled.py:
- Adapter trimmed: neml2/pyzag/operators/_assembly.py now holds only _require_le_one_intmd (the backend envelope guard), with an updated docstring; removed the three helpers from operators/__init__.py.
- Call sites rewired to the ES API
- Tests moved to the ES layer (tests/unit/test_solvers.py)

neml2/pyzag/operators/_flat.py
hdt5kt: - Moved the pure DOF-count helpers onto AxisLayout (neml2/es/axis_layout.py): group_intmd_ndim(), group_flat_size(), flat_size() — docstrings distinguish them from the base-only group_size/storage_size.
- Moved the flat pack/unpack onto AssembledVector (neml2/es/assembled.py): to_flat() and from_flat(layout, flat).
- Deleted neml2/pyzag/operators/_flat.py and dropped its imports/exports from operators/__init__.py.
- Rewired the adapter call sites to the ES API: _vector.py (layout.group_intmd_ndim), _jacobian.py (flat_size(), from_flat), _wrapper.py (to_flat(), from_flat), interface.py (to_flat(), from_flat).
- Add tests

neml2/pyzag/interface.py
hdt5kt: .contiguous() isn't required downstream — J_diag only feeds lu_factor and matmul, which both copy non-contiguous input internally.
-  _expand_am_dynamic now passes full-size blocks.
-  Drop the per-iteration full copy of the diagonal Jacobian in the common case.

 __setitem__ the whole-block clone is copy-on-write safety, aka, blocks are shared via __getitem__ views and the cached factorization, so an in-place slice write could corrupt a parent operator. It's setup-time rather than per-iteration.

hugary1995 (commented):
Re-reviewed at `af9def2a` (rebased onto `b3649cfd`). Both blocking items from the last round are resolved, and I verified the rest independently rather than going by thread state — everything checks out. Thanks for the thorough turnaround.

**Verified:**

- **The `expand_at` fix is properly separated** — merged as #443, and it landed *with* the regression tests (`test_crystal_plasticity.py`, `test_sub_batch_shape.py`, 69 lines). This PR no longer touches `neml2/types/_base.py` at all.
- **Lifted PCR is removed cleanly, not just deleted** — `_lifted_pcr.py` + `test_lifted_pcr.py` gone; the `_select_inverse` / `_has_block_group` dispatch removed from `_jacobian.py` (`inverse_operator` now passes through verbatim); exports cleaned from both `__init__.py`s; `_PCR_MESSAGE` no longer references the deleted class; `CLAUDE.md` / `pyzag.md` / the CP-calibration notebook updated. A BLOCK layout requesting PCR now raises a clear `NotImplementedError` pointing at Thomas — correct degradation.
- **Everything from the prior round survived the rebase**: no `to_torch`, `NEML2PyzagModel` restored, no `compile` constructor param (dynamo bump correctly relocated to `neml2.compile`), no `examples/` under `neml2/`, `LUCache` in `neml2/solvers/`.
- Nice to see `_jacobian.py` actually consuming the promoted ES methods (`sub_am.batch[1:]`, `diag_walk.batch[1:].transpose()`) rather than local helpers — the promotions aren't dead code.
- **Tests: 111 passed** (`tests/unit/pyzag/` + `test_solvers.py` + `test_region_views.py`), run locally against this head.

The PR's contribution vs main is now +2,927/−833 across 41 files, down from +5,172/−1,258. Much more self-contained.

I've resolved the six threads that were still marked open but which I confirmed are fixed (`_cache.py`, `_operator.py` matvec, `_assembly.py`, `_flat.py`, `interface.py` expand, and the deleted calibration example).

---

Two items left — one required, one optional. Details inline.

**Required before merge:** `LUCache` needs a direct unit test (inline on `neml2/solvers/_lu_cache.py`).

**Optional:** stale `factory` wording now that the class is `NEML2PyzagModel` again. Three spots — two flagged inline, plus `doc/content/tutorials/optimization/crystal_plasticity_calibration/main.ipynb` (the `neml2.compile(factory)` cell), which I couldn't anchor inline cleanly in a notebook.

neml2/solvers/_lu_cache.py
```diff
@@ -0,0 +1,74 @@
...
+        self._piv: torch.Tensor | None = None
+
+    def factor(self, matrix: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
+        """Return the cached ``(LU, pivots)`` of ``matrix``, refactoring on change."""
+        if matrix is not self._src or matrix._version != self._version:
+            self._lu, self._piv = self._factor_fn(matrix)
+            self._src = matrix
+            self._version = matrix._version
+        assert self._lu is not None and self._piv is not None
+        return self._lu, self._piv
```
hugary1995, line 64: **Required before merge: this needs a direct unit test.**

`LUCache` is now a shared component in `neml2/solvers/` — reusable by any LU-based solver — but it's only exercised indirectly through `CachingLU` in the pyzag tests. The caching logic is exactly the kind of thing that silently degrades (to "always refactor", or worse, "never refactor") without anyone noticing, since both failure modes still produce correct numbers.

Please add a focused test in `tests/unit/test_solvers.py` covering the invalidation contract:

- same tensor object solved twice → `factor_fn` called **once** (inject a counting `factor_fn` to assert this — that's what the parameter is for);
- in-place mutation of the source (bumping `_version`) → refactors;
- a different tensor object → refactors;
- `invalidate()` → drops the cache and releases the held source.

The first two are the ones that actually protect the optimization and the correctness guarantee respectively.

neml2/pyzag/interface.py
```diff
@@ -22,110 +22,99 @@
...
-    Additional ``args`` and ``kwargs`` are forwarded to
-    :class:`torch.nn.Module` verbatim.
+        exclude_parameters (list of str): NEML2 parameters to *not* mirror as torch
+            parameters. Mutually exclusive with ``include_parameters``.
+        include_parameters (list of str): the *only* NEML2 parameters to mirror as
+            torch parameters. Mutually exclusive with ``exclude_parameters``.
+
+    Construction yields a ``py-eager`` factory. To accelerate the residual with
+    in-process ``torch.compile`` (``py-jit``), opt in explicitly after
+    construction with :func:`neml2.compile` -- e.g. ``neml2.compile(factory)``.
```
hugary1995: **Optional nit:** stale wording — `neml2.compile(factory)` and "yields a `py-eager` factory" read oddly now that the public name is `NEML2PyzagModel` again. Suggest `neml2.compile(model)` (and "a `py-eager` model") for consistency. Same wording appears in `doc/content/deployment/overview.md` and the crystal-plasticity calibration notebook.

doc/content/deployment/overview.md
```diff
@@ -145,7 +145,7 @@ on one of the runtimes above:
   load history on `py-eager`. If the input file names an `AOTIModel`, the same
   driver runs on `py-aoti` instead.
 - The **pyzag** adapter (`NEML2PyzagModel`) drives calibration on `py-eager`,
-  optionally accelerated to `py-jit` via `neml2.compile`.
+  optionally accelerated to `py-jit` via `neml2.compile(factory)`.
```
hugary1995: **Optional nit:** same stale `factory` wording as `interface.py` — `neml2.compile(factory)` → `neml2.compile(model)`, matching the restored `NEML2PyzagModel` name.

PR #447, opened 2026-08-12
Coordinate-descent predictor: 16 -> 5 on crystal plasticity, 15 -> 0 on viscoplasticity

hugary1995:
Warm-starts an implicit constitutive update by coordinate descent on the
condensed rate system, adds the runtime machinery the compiled routes need to
run it, and retires the alternative model form that existed to avoid needing it.

| | cold step | total |
| --- | --- | --- |
| crystal plasticity (8 scenarios) | 16 -> **5**, 14 -> **3**, 10 -> **4** | -1.9% to -3.6% |
| viscoplasticity (4 of 5) | 15 -> **0** | 61 -> 46, 129 -> 114, 128 -> 113 |
| viscoplasticity (chaboche) | 15 -> **3** | 310 -> 298 |

Zero means Newton returns *at the predictor*: the guess already satisfies the
residual to `atol`. That holds at every step size across six decades of `dt`,
which is the part that matters — see "Retiring the inverted form" below.

## Why coordinate descent

Condensed onto the rates the system is `phi(gdot) + A gdot = b`, which is the
gradient of a strictly convex potential — `phi` is monotone and `A` is positive
semi-definite. Coordinate descent on that is globally convergent with no
globalization, and each coordinate step is an exact scalar solve rather than a
linearization. Newton, by contrast, crawls from a near-zero start because `phi'`
is unbounded at the origin for a rate-sensitive law.

The only physics hook is the explicit rate law itself — driving force in, rate
out. Every rate-dependent physics already has one.

## Where `A` comes from

Two routes, because the right answer differs by physics.

**Closed form, for crystal plasticity.** `SlipSystemElasticInteraction` gives
`A = dt M^T C M` in one block. It drops the spin convection deliberately: the
exact condensed coupling is `dt M^T C (I - dt Omega)^-1 M`, which is not
symmetric and so is not the gradient of anything, and coordinate descent needs
symmetry. Omega is skew, so its diagonal contribution is exactly zero and the
`A_ii >= 0` bracket precondition survives untouched.

**Differentiated, for viscoplasticity.** `RateCondensation` reads
`b = f(0)` and `A = -f'(0)` off a single `Model.jvp` of the authored return
path — one forward pass with a tangent riding the existing `forward(v=)` chain
rule, so neither a finite difference nor reverse-mode autograd. It needs only
the first-order chain rule every leaf already has, which is why it costs no
per-physics code.

Viscoplasticity needs the second route because writing `A` out means one term
per hardening mechanism:

```
A = dt [ N:C:N  +  h'(ep)  +  sum_i N : dXdot_i/dgammadot ]
```

and **lagging the hardening does not work**. Freezing it leaves `A = dt N:C:N`,
whose relative error in `dgamma` is `h'/(N:C:N)` — under a percent for the usual
metal parameterizations. But the returned stress is a near-cancellation of the
trial stress against the plastic relaxation, and amplifies that error by
`sigma_trial/sigma`, which grows with the step. Measured: lagging halves the
cold step below `dt ~ 1` and buys nothing above it, and on chaboche at the two
largest steps it is *worse than no predictor at all*, because the guess lands
where the yield function has gone negative.

| cold step, `end_time` | 0.1 | 1 | 10 | 100 | 1e3 | 1e4 | 1e5 |
| --- | --- | --- | --- | --- | --- | --- | --- |
| no predictor | 5 | 6 | 8 | 10 | 11 | 13 | 15 |
| lagged coupling | 2 | 3 | 3 | 4 | 6 | 13 | 15 |
| **differentiated coupling** | **0** | **0** | **0** | **0** | **0** | **0** | **0** |
| inverted formulation | 9 | 8 | 7 | 6 | 7 | 6 | 6 |

(`isoharden`; the other linear-hardening scenarios are identical. chaboche is
2–3 across the row, because Voce plus Chaboche recovery make the return path
genuinely nonlinear over a step.)

Crystal plasticity tolerates lagging for a reason that is worth stating: there
the hardening enters `tauhat` **inside phi**, not the driving force, so freezing
it perturbs `phi` and never `A`. In viscoplasticity the hardening is *in* `f`,
so freezing it perturbs `A` directly — and `A` is what the cancellation
amplifies. Same approximation, opposite consequence.

## Retiring the inverted form

`PerzynaPlasticFlowRateResidual` stated the flow rule as
`eta*gdot^(1/n) - <f>` with the flow rate carried as an extra unknown. It
existed because Newton converged from a cold start in 9 iterations against the
rate form's 15 and, unlike the rate form, did not degrade as the step grew.

The predictor now beats it on both counts (table above), so the alternative
model form is pure debt: a second way to write the same physics, with its own
regularization branch, cutoff parameter, two `ModelUnitTest` fixtures and three
regression scenarios, all maintained so a predictor could be avoided — and all
of it kept correct across six evaluation routes.

**This is an API break.** An input file naming `PerzynaPlasticFlowRateResidual`
now fails at load with "not registered in NativeRegistry".
`PerzynaPlasticFlowRate`, the explicit form every scenario actually uses, is
untouched.

## Also a behaviour change

`ImplicitUpdate` used to catch `KeyError` around predictor construction and fall
back to no predictor. It caught the error from *anywhere* inside constructing
the predictor, so a stale model name in a `ComposedModel` became a warning plus
a silently disabled predictor. Six scenarios in this branch appeared to diverge
under it while actually running with no predictor at all. The catch is gone: an
input file that previously limped along now fails loudly.

## What is new

**Models**
- `CoordinateDescentPredictor` — Gauss-Seidel over the components, each
  coordinate bisected in the driving-force variable, where the map has unit
  slope at the origin instead of being vertical. Handles both the vector layout
  (`A` over `(m, m)`, `b` over `(m,)`) and the degenerate one-coordinate layout
  (plain scalars) that viscoplasticity needs.
- `SlipSystemElasticInteraction` — the crystal-plasticity coupling matrix.
- `RateCondensation` — `b` and `A` from one jvp of the return path.
- `ConstantExtrapolationPredictor` / `LinearExtrapolationPredictor` gain a
  `cold = 'unknown:variable'` mapping: the coordinate-descent guess on the step
  with no history to extrapolate from, ordinary extrapolation after. The cold
  test is the extrapolator's own `|t~1 - t~2| <= eps`, which is the actual
  question — a test on a variable's magnitude only coincides with coldness when
  that variable starts at zero, false for a dislocation density or an identity
  `Fp`.
- `ForwardEulerTimeIntegration` gains an `old_variable` rename knob, symmetric
  with the `rate` knob it already had. Without it a provisional output could
  only ever start from a previous step of *itself*.

**Runtime loop (`IterableExport`, AOTI schema v14)**

The predictor is a bounded iteration, so exporting it as written would unroll
thousands of copies of the inner rate law into the graph and freeze the sweep
count at compile time. Instead a model may implement `iterable_export_form()`:
the exporter compiles a *single* step taking the previous value on an extra
input, and the C++ runtime re-runs the graph feeding the output back — the same
shape as the Newton loop beside it. Nothing in the exporter or runtime knows
what coordinate descent is, and the count lives in metadata so it is retunable
without recompiling. One C++ loop serves `py-aoti`, `cpp-aoti` and
`cpp-dispatch`; the three eager routes run the Python model's own loop.

**Companions**

Thirteen `*_cd_predictor` scenarios, each beside its parent with the parent's
gold **copied unchanged** and everything above the final `[Models]` block
byte-identical. That is the point: a predictor moves only the initial guess, so
passing against the parent's reference proves the coupling does not perturb the
converged answer.

The predictor graphs list the residual's *own* blocks wherever a variable name
carries the same value in both graphs, rather than duplicating them. What turns
that on is emitting the frozen directions and the rotation matrix under their
plain names. What stays duplicated is what genuinely differs: the trial chain
(same leaves, trial values), the lagged strength/hardening leaves (they read
`~1`), and the forward-Euler integrators — the residual integrates backward,
which is a different model rather than a rename.

**A guard for a bug the golds cannot see**

`tests/regression/test_predictor_wiring.py`. A predictor consumes old state and
driving forces, nothing else; ask for more and `TransientDriver` zero-fills it —
deliberately, since that is how initial conditions work — and the predictor
quietly returns a degraded guess while every gold still passes. That produced
two false negatives here: a predictor asking for `deformation_rate` in a
scenario prescribing `spatial_velocity_gradient`, and a reused block asking for
`R` in a scenario that renames its rotation matrix. Both measured as "the
predictor does nothing" rather than as an error. The invariant is static and
cheap: 68 scenarios in half a second, and verified to fail on a surgical
reintroduction of the second bug.

## Verification

- `tests/unit` 871, `tests/models` + `tests/regression` 315 — all green.
- Every companion reproduces its parent's gold unchanged.
- `neml2-compile` builds `model_predictor.pt2` for a `RateCondensation`
  predictor — the nested `Model.jvp` inside `forward` does lower — and
  `py-aoti == py-eager` to 3e-14 on a batched call.
- The C++ runtime loop was confirmed to be doing the iterating by lowering
  `iterations` in metadata and watching the Newton count move on the same graph.

## Follow-ups, not in this PR

- **#448** — the chain rule cannot seed a derivative dense across a sub-batch
  axis, which is what blocks moving crystal plasticity onto `RateCondensation`
  as well. Enabler only: for the small-strain scenarios both routes give the
  same `A`, so the closed form stays correct.
- **Globalization.** Rate + coordinate descent reaches x4 step size against the
  inverted form's x20. Closing that needs a globalization bounding the
  per-iteration growth of `tau/tauhat`, which is a separate investigation.


neml2/models/common/RateCondensation.py
```diff
@@ -0,0 +1,180 @@
...
+        # A scalar zero (and unit tangent) with no batch of its own: the batch
+        # broadcasts up from the path's other inputs, so this leaf needs no way
+        # to ask what the rate's batch shape is. Sound for a scalar rate, which
+        # is the one-coordinate condensation. A rate carried on a sub-batch axis
+        # (crystal plasticity) would need a seed that is an identity over that
+        # axis -- one pass with a K-wide tangent, but not this seed.
+        ref = next(iter(pt.values()))
+        rate_cls = self.model.input_spec[self._rate]
+        zero = rate_cls(torch.zeros((), dtype=ref.dtype, device=ref.device))  # type: ignore[call-arg]
+        one = rate_cls(torch.ones((), dtype=ref.dtype, device=ref.device))  # type: ignore[call-arg]
```
hugary1995, line 187: **A non-scalar `rate_cls` corrupts the seed silently.** The docstring says this is "sound for a scalar rate", but nothing enforces it, and the typed wrappers do not object:

```
>>> SR2(torch.zeros(())).data.shape
torch.Size([])          # against SR2.BASE_SHAPE == (6,)
>>> SR2(torch.zeros(())).batch_shape
torch.Size([])
```

No exception — just a malformed wrapper that goes into `self.model.jvp` and produces garbage or a confusing error somewhere downstream. Since `rate` is a free-text HIT option naming any input of the path model, this is reachable by a wiring mistake rather than only by a code change.

Two lines in `__init__`, beside the two checks already there:

```python
if model.input_spec[rate].BASE_SHAPE != ():
    raise ValueError(
        f"RateCondensation: rate {rate!r} is "
        f"{model.input_spec[rate].__name__}; the zero-rate seed is only "
        "defined for a scalar rate."
    )
```

neml2/models/common/CoordinateDescentPredictor.py
```diff
@@ -0,0 +1,449 @@
...
+        # Newton polish off the rate law's own tangent:
+        # d/dw [w + aii*phi_inv(w)] = 1 + aii * dphi_inv/dw. Safe here because
+        # bisection has already put the iterate at the root.
+        for _ in range(self.polish):
+            _, jvp = self.rate_law.jvp(
+                {self.driving_force_input: w, **pt},
+                {self.driving_force_input: w * 0.0 + 1.0},
+            )
+            dg = next(iter(jvp.values()))
+            g = self._phi_inv(w, pt)
```
hugary1995: **The polish step throws away a rate-law evaluation it has already paid for.** `jvp` returns `(out, jvp_out)`, and `out` at this `w` is exactly the `g` recomputed on line 277.

Measured by instrumenting `PowerLawSlipRule.__call__` (m=3, sweeps=2, defaults otherwise): **23 rate-law forwards per coordinate solve — 16 bisections + 3 inside `jvp` + 3 redundant + 1 final.** With

```python
out, jvp = self.rate_law.jvp(...)
dg = next(iter(jvp.values()))
g = next(iter(out.values()))
```

it drops to **20 (−13%)**, `tests/unit/test_coordinate_descent.py` stays green and the results are unchanged. At the CP defaults (12 systems × 16 sweeps) that is 576 wasted rate-law evaluations per predictor call, which is worth having given the reported CP total is −1.9% to −3.6%.

While here: `bisections` defaults to 16, but its own schema docstring says "unchanged from 6 upward and broken at 4". Eight would keep 2× margin and remove another 8 of the 20 — 3 Newton steps still take a 2⁻⁸ bracket to ~1e-19 quadratically. You have the measurement rig; the two changes together roughly halve the predictor's cost.

neml2/models/common/CoordinateDescentPredictor.py
```diff
@@ -0,0 +1,449 @@
...
+        return self._phi_inv(w, pt)
+
+    def _slice_pt(self, pt: dict[str, Scalar], i: int, m: int) -> dict[str, Scalar]:
+        """The rate law's other inputs at component *i*.
+
+        A per-component input (the slip strengths) is indexed; anything carried
+        once for the whole vector is passed through untouched.
+        """
+        return {
+            n: (v.sub_batch[i] if v.sub_batch_shape and v.sub_batch_shape[-1] == m else v)
```
hugary1995: **The predicate and the action disagree about which axis.** The test reads the *last* sub-batch axis (`sub_batch_shape[-1]`), but `SubBatchView.__getitem__` documents `V.sub_batch[0]` as picking the *first* sub-batch site. Those coincide only at sub-batch rank 1.

So a passthrough carried at sub-batch shape `(k, m)` passes the test and is then sliced along the size-`k` axis — the wrong value, or an `IndexError` if `k < m`. Today every rate law's passthrough is rank-1 (`slip_strengths`), so this is latent rather than live, but it is the sort of thing that reads as a plausible number rather than as a failure.

Tightening the predicate to match the action costs nothing:

```python
n: (v.sub_batch[i] if v.sub_batch_ndim == 1 and v.sub_batch_shape[0] == m else v)
```

(`sub_batch_labels` would give exact axis identity, but it does not exist in v3 — `CLAUDE.md` is stale on that point.)

neml2/models/common/CoordinateDescentPredictor.py
```diff
@@ -0,0 +1,449 @@
...
+    def _check_layout(A: Scalar, b: Scalar) -> None:
+        """Reject a mismatched ``(A, b)`` pair before it reads as garbage.
+
+        Sub-batch rank is the only thing distinguishing the two layouts, so a
+        mis-wired input -- an ``A`` that forgot its second axis, say -- would
+        otherwise index into the wrong axis and silently produce a plausible
+        number.
+        """
+        want = 0 if b.sub_batch_ndim == 0 else 2
+        if A.sub_batch_ndim != want or b.sub_batch_ndim > 1:
```
hugary1995, line 353: **Ranks are checked, extents are not** — so the case the docstring names is only half-covered. "An `A` that forgot its second axis" is caught, but an `A` at `(m, k)` with `k != m` passes rank 2 and then `_entry` reads entries that are not there (or `IndexError`s), which is the same "plausible number" outcome the check exists to prevent.

One more assertion closes it:

```python
if b.sub_batch_ndim == 1:
    m = int(b.sub_batch_shape[0])
    if tuple(A.sub_batch_shape) != (m, m):
        raise ValueError(
            f"CoordinateDescentPredictor: coupling is sub-batched "
            f"{tuple(A.sub_batch_shape)}, expected ({m}, {m}) to match the "
            f"driving force."
        )
```

Neither this `ValueError` nor the "rate law must have exactly one output" one in `__init__` is currently exercised by a test.

neml2/cli/aoti_export.py
```diff
@@ -311,6 +311,82 @@ def _segment_var_infos(spec_or_infos) -> list[dict]:
...
+    than silently looping just the first.
+
+    A predictor with no iterable leaf takes the original path unchanged --
+    same wrapper, same graph, same metadata -- so nothing here costs an
+    existing predictor anything at compile time or at run time.
+    """
+    from ..models.common import ComposedModel  # noqa: PLC0415
+
+    leaves = list(_flatten_composed(pred))
+    found = [(i, m) for i, m in enumerate(leaves) if hasattr(m, "iterable_export_form")]
```
hugary1995: **The `IterableExport` opt-in is pure duck-typing with nothing declared on `Model`.** `hasattr(m, "iterable_export_form")` is the entire protocol, and the only place a future author would learn it exists is by reading this function or `CoordinateDescentPredictor`. `IterableExport` itself already lives in `model.py` — the hook belongs next to it, either as a `typing.Protocol` the exporter narrows against or as a documented optional method on `Model` (a no-op default returning `None`, with the check becoming `leaf.iterable_export_form() is not None`).

Minor, and in the same area: the metadata pair is `predictor_feedback.name` for the input but `.output` for the output. Since this is a persisted versioned format, `input`/`output` is cheaper to make symmetric now than at v15.

hugary1995 (commented):
Review notes from a read-through plus a local run of the Python routes.

**Verified locally** (worktree at `8f2ea20c`, py-eager only): `tests/unit` + `tests/models` 1048 passed; the four new test modules 84 passed; two CP `*_cd_predictor` regression scenarios passed; `ruff check`/`format` clean; `pyright` shows no new errors in changed files; all 13 companion golds are byte-identical to their parents (`cmp`); `neml2-syntax` lists the four new types and no longer lists `PerzynaPlasticFlowRateResidual`, with no dangling references anywhere in the tree.

**Not verified:** `tests/aoti/` and the C++ ctests — my environment's `libneml2.so` is a schema-v13 build, so the `predictor_feedback` contract and `_run_predictor` are entirely on your run.

Five comments inline. Two are silent-corruption paths that a couple of lines convert into a clear error, one is free performance I measured, and two are smaller.

Separately, a testing note that doesn't fit a line anchor: **`iterable_export_form` equivalence has no unit test.** I checked locally that N applications of `_forward_one_sweep` equal `forward(sweeps=N)` exactly (max diff `0.0`), so the mechanism is correct today. It's worth pinning anyway, because the mechanism is an instance-level `forward` monkeypatch on a `copy.copy`'d module and the AOTI suite cannot catch it silently ceasing to take effect: 16 unrolled sweeps looped 16 times converges to the same root, and `test_the_runtime_really_iterates` would still see the Newton count move. Fifteen lines in `test_coordinate_descent.py` closes it.

PR #456, opened 2026-08-31
Make c10::LinAlgError recoverable at every NEML2 C++ boundary

reverendbedford:
A singular / non-invertible / non-PD factorization from a torch linalg kernel is numerically recoverable in the same sense as a Newton non- convergence: a downstream time-stepping consumer can cut the step and retry. It was previously falling through the generic std::exception catch in the aoti and eager guards and surfacing as a non-recoverable FatalError, so callers branching on `recoverable()` were forced to give up on a failure they could have driven around. The original torch message is carried through on the re-thrown ConvergenceError so the end user still sees which factorization failed.

### DIFF
--- a/neml2/csrc/aoti/assertions.h
+++ b/neml2/csrc/aoti/assertions.h
@@ -27,10 +27,11 @@
 #include <sstream>
 
 // c10::LinAlgError is a numerical failure raised by torch's linalg kernels
-// (a singular / non-invertible / non-PD factorization); a downstream time-
-// stepping consumer can recover from it by cutting the step, so the guard below
-// re-throws it as the recoverable neml2 ConvergenceError rather than letting it
-// fall through to the generic std::exception catch (which would flag it fatal).
+// (e.g., a singular / non-invertible or non-positive-definite matrix); a
+// downstream time-stepping consumer can recover from it by cutting the step, so
+// the guard below re-throws it as the recoverable neml2 ConvergenceError rather
+// than letting it fall through to the generic std::exception catch (which would
+// flag it fatal).
 #include <c10/util/Exception.h>
 
 #include "neml2/csrc/aoti/Exception.h"
@@ -73,9 +74,10 @@ _assert(bool cond, const Args &... args)
 /// neml2 exception taxonomy so a consumer has a single catch surface:
 ///   - a neml2 `Exception` (incl. the recoverable `ConvergenceError`) passes
 ///     through untouched -- its dynamic type and `recoverable()` are preserved;
-///   - a torch `c10::LinAlgError` (a singular / non-invertible / non-PD linear
-///     factorization from a torch linalg kernel) is re-thrown as the recoverable
-///     `ConvergenceError` -- like a Newton non-convergence, a time-stepping
+///   - a torch `c10::LinAlgError` (e.g., a singular / non-invertible or
+///     non-positive-definite matrix from a torch linalg kernel) is re-thrown as
+///     the recoverable `ConvergenceError` -- like a Newton non-convergence,
+///     a time-stepping
 ///     consumer can cut the step and retry. Recognized BOTH as the typed C++
 ///     exception (when RTTI matches across libraries) AND by the preserved
 ///     "torch.linalg.<op>:" message prefix on a plain `std::exception` (torch's
--- a/neml2/csrc/aoti/assertions.h
+++ b/neml2/csrc/aoti/assertions.h
@@ -26,6 +26,13 @@
 
 #include <sstream>
 
+// c10::LinAlgError is a numerical failure raised by torch's linalg kernels
+// (a singular / non-invertible / non-PD factorization); a downstream time-
+// stepping consumer can recover from it by cutting the step, so the guard below
+// re-throws it as the recoverable neml2 ConvergenceError rather than letting it
+// fall through to the generic std::exception catch (which would flag it fatal).
+#include <c10/util/Exception.h>
+
 #include "neml2/csrc/aoti/Exception.h"
 
 // Internal header -- NOT shipped. Holds the check-and-throw helpers shared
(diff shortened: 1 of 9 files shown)


neml2/csrc/aoti/assertions.h
```diff
@@ -26,6 +26,13 @@
 
 #include <sstream>
 
+// c10::LinAlgError is a numerical failure raised by torch's linalg kernels
+// (a singular / non-invertible / non-PD factorization); a downstream time-
```
ChengHauYang: (a singular/non-invertible or non-positive-definite matrix)

neml2/csrc/eager/Model.cpp
```diff
@@ -61,6 +61,30 @@ is_convergence_error(const py::error_already_set & e)
   }
 }
 
+// True if the Python exception `e` is (or derives from) `torch.linalg.LinAlgError`
+// -- the Python face of a c10::LinAlgError raised by a torch linalg kernel
+// (singular / non-invertible / non-PD factorization). Numerically recoverable in
```
ChengHauYang: (e.g., a singular/non-invertible or non-positive-definite matrix)

tests/cpp/fixtures/linalg_error_model.py
```diff
@@ -0,0 +1,70 @@
...
+        input("in_stress", SR2, "Input stress (ignored)"),
+        output("out_stress", SR2, "Output stress (unreachable)"),
+    )
+
+    def forward(  # type: ignore[override]
+        self,
+        in_stress: SR2,
+        v: ChainRuleDict | None = None,
+    ) -> SR2 | tuple[SR2, ChainRuleDict]:
+        # Build a zero PSD matrix on the same device/dtype as the input, then
```
ChengHauYang: Build a zero positive-semidefinite matrix

tests/cpp/test_exceptions.cpp
```diff
@@ -139,6 +144,64 @@ main()
   }
   NEML2_CHECK_THROWS(_assert(false, "nope"));
 
+  // --- _guarded normalizes c10::LinAlgError -> recoverable ConvergenceError --
+  // A torch linalg kernel that hits a singular / non-PD matrix throws
```
ChengHauYang: singular / non-positive-definite

tests/cpp/fixtures/linalg_error_model.py
```diff
@@ -0,0 +1,70 @@
...
+        # ask torch for its Cholesky factor. Zero has no positive-definite
```
ChengHauYang: The zero matrix is not positive-definite, so the Cholesky factorization fails.

tests/cpp/test_exceptions.cpp
```diff
@@ -139,6 +144,64 @@ main()
...
+  // c10::LinAlgError; the guard must re-throw it as our recoverable
+  // ConvergenceError so a time-stepping consumer cuts the step and retries
+  // instead of hard-failing. The original message must survive so the end user
+  // can still see which factorization failed.
+  {
+    bool threw = false, recoverable = false, kept_msg = false, is_fatal = false;
```
ChengHauYang: I suggest changing `threw` to `caught_convergence_error`.

hdt5kt:
The fixes make sense, and I think the missing code coverage is too specific (hard to trigger) so I think it is fine.

neml2/csrc/eager/Model.cpp
```diff
@@ -89,10 +113,17 @@ guarded(const char * op, F && f) -> decltype(f())
...
+    // it as the recoverable C++ type so a host can cut the step and retry. A
+    // torch.linalg.LinAlgError from a torch linalg kernel (a c10::LinAlgError
+    // that crossed into Python) is likewise numerically recoverable and gets the
+    // same treatment. Everything else is a fatal config / shape / dtype error.
+    // (Split into two separate ifs rather than `||` so branch-coverage tools
+    // don't flag one side of the short-circuit as partial when the other is
+    // hit -- both recoverable cases are tested independently below.)
     if (is_convergence_error(e))
       throw neml2::aoti::ConvergenceError(prefix + e.what());
+    if (is_torch_linalg_error(e))
```
ChengHauYang: `guarded()` is also used by the Model constructor below, so this would classify any torch.linalg.LinAlgError raised during model loading or construction as recoverable. For example, a failure caused by an invalid parameter is unlikely to change when the time step is reduced, whereas a singular Jacobian or residual evaluation may succeed after retrying with a smaller step. Should `guarded()` take an explicit policy so construction-time `LinAlgError`s remain fatal while evaluation-time failures are recoverable? A construction-time test would help clarify the intended behavior.

neml2/csrc/aoti/assertions.h
reverendbedford: Agreed

neml2/csrc/eager/Model.cpp
reverendbedford: Agreed

tests/cpp/fixtures/linalg_error_model.py
reverendbedford: Agreed

tests/cpp/test_exceptions.cpp
reverendbedford: Agreed

tests/cpp/fixtures/linalg_error_model.py
reverendbedford: Agreed

tests/cpp/test_exceptions.cpp
reverendbedford: Agreed

neml2/csrc/eager/Model.cpp
reverendbedford: This is a good catch, changed things so that we still raise construction time errors.

neml2/csrc/aoti/assertions.h
```diff
@@ -76,8 +76,13 @@ _assert(bool cond, const Args &... args)
 ///   - a torch `c10::LinAlgError` (a singular / non-invertible / non-PD linear
 ///     factorization from a torch linalg kernel) is re-thrown as the recoverable
 ///     `ConvergenceError` -- like a Newton non-convergence, a time-stepping
-///     consumer can cut the step and retry. The original message is carried on
-///     the new exception so the end user still sees which factorization failed;
+///     consumer can cut the step and retry. Recognized BOTH as the typed C++
+///     exception (when RTTI matches across libraries) AND by the preserved
```
ChengHauYang: exception when run-time type information (RTTI) matches across libraries,

tests/cpp/test_exceptions.cpp
```diff
@@ -202,6 +202,34 @@ main()
...
     NEML2_CHECK(fatal);
   }
+  // Message-based fallback: the AOTI proxy_executor C API strips the derived
+  // c10::LinAlgError type when a linalg kernel inside an AOT-compiled graph
+  // throws, so what escapes the compiled .so is a plain std::exception that
+  // carries the original "torch.linalg.<op>:" message but not the LinAlgError
+  // type. Simulate that here -- the guard must still promote it to a recoverable
+  // ConvergenceError based on the message prefix.
+  {
+    bool got_conv = false, recoverable = false, kept_msg = false;
```
ChengHauYang: suggestion: `got_conv` -> `caught_convergence_error`

ChengHauYang (approved):
Thanks, Mark! It looks good to me!

PR #458, opened 2026-09-09
Materialize broadcast initial guesses in Newton::solve_masked

ChengHauYang:
solve_masked promises that `u` and `converged_mask` index the same rows, but an unbatched initial guess only picked up the residual's dynamic batch as a side effect of a Newton update. With zero iterations (max_its = 0, or a first-iteration bail) the best-effort iterates came back unbatched beside a batched mask, so a consumer zipping the two got a shape mismatch.

Expand each unknown group to the dynamic batch on entry, before the iteration loop, and document the invariant on `NewtonResult::u`. The throwing `solve` path is unaffected: it leaves `converged_mask` undefined and makes no such promise.

Covered by a new zero-iteration case in the failure-capture suite (`max_its` is now a knob on the generated input) asserting both the mask and the captured unknowns carry the 2x2 batch.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -60,7 +60,10 @@ jobs:
         # and only needed when the validator runs, so it's fetched here
         # rather than at wheel install time. The validator's error list
         # lands in the job step summary via $GITHUB_STEP_SUMMARY.
-        run: playwright install --with-deps chromium
+        run: |
+          # Remove the unrelated Chrome repository to keep CI stable.
+          sudo rm -f /etc/apt/sources.list.d/google-chrome.list
+          playwright install --with-deps chromium
       - name: Validate built doc site (MathJax, JS errors, broken assets)
         run: python doc/scripts/check_doc_render.py doc/_build/html
       - name: Write build metadata for the deploy job
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -60,10 +60,7 @@ jobs:
         # and only needed when the validator runs, so it's fetched here
         # rather than at wheel install time. The validator's error list
         # lands in the job step summary via $GITHUB_STEP_SUMMARY.
-        run: |
-          # Remove the unrelated Chrome repository to keep CI stable.
-          sudo rm -f /etc/apt/sources.list.d/google-chrome.list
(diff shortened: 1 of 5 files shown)


.github/workflows/docs.yaml
```diff
@@ -60,7 +60,10 @@ jobs:
         # and only needed when the validator runs, so it's fetched here
         # rather than at wheel install time. The validator's error list
         # lands in the job step summary via $GITHUB_STEP_SUMMARY.
-        run: playwright install --with-deps chromium
+        run: |
+          # Remove the unrelated Chrome repository to keep CI stable.
+          sudo rm -f /etc/apt/sources.list.d/google-chrome.list
```
reverendbedford: There's no way this is going to remain stable as github changes runners, right?

neml2/aoti/__init__.py
reverendbedford, line 1: How is this related to the rest of the change?

reverendbedford (changes requested):
The main change is fine, but what about the random other stuff?

.github/workflows/docs.yaml
ChengHauYang: Sure! Thanks, @reverendbedford! I reverted it.

neml2/aoti/__init__.py
ChengHauYang: Hi @reverendbedford, Thanks for your feedback!                                                                                                                                         

This is related to the new `test_py_eager_zero_iteration_capture_enriches_error` test. The test directly accesses `err.converged_mask` and `err.unknowns`. The test itself runs correctly, but when CI runs **Pyright**, Pyright only sees the auto-generated `_aoti.pyi`, which does not describe these two attributes, so the new test causes the type check to fail.

This patch gives Pyright the complete type description from `neml2.solvers._exceptions` through the `TYPE_CHECKING` branch. At runtime, it still imports the original `ConvergenceError` from `._aoti`, so there is no runtime behavior change. **Without this patch, the test would pass, but the Pyright check in CI would fail**.

ChengHauYang:
Hi @reverendbedford, thanks for your feedback. I reverted the documentation changes. The remaining change in `aoti/__init__.py` is needed for type checking with Pyright in CI.

reverendbedford (approved):

PR #461, opened 2026-09-16
Generalize device handling to any torch accelerator (XPU, HIP, MPS)

reverendbedford:
Every NEML2 code path -- the AOTI compile CLI, the on-disk artifact layout, the Python AOTI shim, the C++ AOTI loader, and the C++ work schedulers -- assumed the target device family was either CPU or CUDA. That assumption was baked in at the argparse `choices=["cpu","cuda"]` enums, the C++ `device_type_str` helper that collapsed anything non-CUDA to `"cpu"` when picking the `<device>/<dtype>/` artifact leaf, the CUDA- only `nvcc` preflight, the loader's CUDA-gated device-index check, and the MPI scheduler's hard `is_cpu() || is_cuda()` gate. Generalize to support any torch accelerator family -- Intel XPU today, plus HIP and MPS as first-class entries -- without changing any behavior for existing CPU or CUDA users.

The runtime turns out to be almost entirely device-opaque already: `at::Device` / `torch.device` are threaded through by value everywhere, `libneml2.so` links only `torch::core` (never `torch::cuda`), and the metadata written by the compiler is already device- and dtype- independent. Adding a family is now a one-line append to the Python `KNOWN_FAMILIES` tuple + one enum in the C++ `is_accelerator()` predicate; every other layer routes the device opaquely.

Python side:
- New `neml2/_accelerator.py` is the single source of truth for `KNOWN_FAMILIES = ("cpu","cuda","xpu","hip","mps")`, `parse_device_spec`, `folder_name` (the lowercase `torch.device.type` string that names the artifact leaf), `target_accelerator_family` (walks example tensors), and a per-family `TOOLCHAIN_CHECKS` table. The CUDA `nvcc` preflight is moved verbatim from `models/export.py`; XPU/HIP/MPS entries start as minimal `torch.<x>.is_available()` probes with `TODO(*-toolchain)` markers for someone with hardware to firm up.
- `neml2/cli/aoti_compile.py`, `neml2/cli/run.py`, `neml2/cli/_compile_wizard.py` widen `--device` from `choices=["cpu","cuda"]` to `list(KNOWN_FAMILIES)`. `neml2-compile` keeps bare family names only (one artifact per family); `neml2-run` additionally accepts indexed forms (`cuda:1`, `xpu:0`) since it sets `torch.set_default_device`. Defaults stay `cpu` / `float64`. The parallel-jobs warning generalizes from "cuda spawns worker contexts" to "any accelerator spawns worker contexts".
- `neml2/aoti/_shim.py` validates the incoming device family up-front through `parse_device_spec` so an unknown family fails with a clean error rather than silently landing in a wrong folder.

C++ side:
- New `neml2/csrc/aoti/DeviceLayout.h` provides `device_folder_name` (delegating to `c10::DeviceTypeName(dev.type(), /*lower_case=*/true)`) and the `is_accelerator()` predicate, mirroring the Python side. `c10::DeviceTypeName(kCPU/kCUDA, true)` returns exactly `"cpu"`/`"cuda"`, so the folder-name mapping is byte-identical to the previous `d.is_cuda() ? "cuda" : "cpu"` on those types and existing `.pt2` bundles keep loading unchanged.
- `neml2/csrc/aoti/Model.cpp` routes through the new helper; the CUDA-only guard on the loader's device index is dropped in favor of `has_index()` alone, which is the correct predicate for any indexed accelerator.
- `neml2/csrc/dispatchers/MPISimpleScheduler.cpp` replaces the CUDA-specific `nbare_cuda`/`cuda_indices` bookkeeping with a per-family `unordered_map<at::DeviceType, ...>`. The rules generalize: at most one bare entry per family, unique indices within a family, and no mixing bare with pinned within a family. Cross-family mixes like `{"cpu","cuda:0","xpu:0"}` are legal.

Build / tests:
- `find_package(torch ...)` gains an optional `xpu` component alongside `cuda`; `Findtorch.cmake` defines `torch::xpu` when `c10_xpu` + `torch_xpu` are present. `tests/cpp/CMakeLists.txt` opts the fixture into an XPU compile when `TARGET torch::xpu`, and `test_dispatcher.cpp` gains a minimum-wire XPU load+forward block gated on `at::hasXPU()`. libneml2.so itself still links only `torch::core` -- accelerator runtimes are picked up transparently via torch's dispatch at load time.
- `tests/unit/test_accelerator.py` covers `parse_device_spec`, `folder_name`, `target_accelerator_family`, and every `TOOLCHAIN_CHECKS` entry (with `torch.<x>.is_available` mocked); runs on any host regardless of installed torch backends.
- `tests/unit/test_cli_device_parsing.py` covers the argparse layer of `neml2-compile` and `neml2-run` plus the shim's family validation.
- `tests/cpp/test_scheduler.cpp` extends `parse_mpi_devices` coverage to XPU/HIP/MPS strings and to cross-family mixes -- all constructible without accelerator hardware because the scheduler only parses device strings via `at::Device`.

Notes on ROCm and XPU / HIP / MPS validation:
- ROCm-built PyTorch already aliases `torch.cuda` to HIP, so AMD GPUs under ROCm wheels are addressed as `"cuda"` and continue to work via the existing CUDA path -- no user-visible change. `"hip"` in the accepted set is for the newer, distinct `torch.hip` / `at::kHIP` surface.
- The XPU / HIP / MPS paths are *wired* -- everything routes through the same generic runtime -- but end-to-end verification on that hardware still needs to happen. A new `doc/content/references/ accelerators.md` page documents the per-family requirements and flags this explicitly.


neml2/_accelerator.py
```diff
@@ -0,0 +1,236 @@
...
+def target_accelerator_family(tensors: Iterable[torch.Tensor]) -> str | None:
+    """Return the first non-``cpu`` ``device.type`` seen across *tensors*,
+    or ``None`` if every tensor is on CPU.
+
+    The AOTI compile preflight uses this to decide which per-family
+    toolchain check to run (:func:`check_toolchain`).
+    """
+    for t in tensors:
+        fam = t.device.type
+        if fam != "cpu":
```
ChengHauYang: how about adding `_NON_ACCELERATOR_FAMILIES = ("cpu", "meta")`?

`meta` is also just for testing shape and it is not an actual accelerator (device) -- same as cpu, it needs no toolchain check. So the flow becomes:

- if fam in _NON_ACCELERATOR_FAMILIES (cpu/meta) -> continue, no toolchain needed
- if fam in KNOWN_FAMILIES (cuda/xpu/hip/mps) -> return fam, run check_toolchain
- if fam not in both lists -> error out. Don't compile silently for a device NEML2's C++ runtime won't route (nobody to look it in the folder, dispatcher treats it as cpu)

Because right now if fam != "cpu": lets `meta` slip through and then check_toolchain("meta") crashes with "no toolchain preflight registered" -- but `meta` is not a real device, it doesn't need one. Unknown family should fail loudly at the front, not confusingly at the end.

tests/unit/test_accelerator.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def test_mps_raises_when_unavailable(self):
+        fake_mps = SimpleNamespace(is_available=lambda: False)
+        fake_backends = SimpleNamespace(mps=fake_mps)
+        with patch("neml2._accelerator.torch", SimpleNamespace(backends=fake_backends)):
+            with pytest.raises(RuntimeError, match="MPS AOTI export requires"):
+                check_toolchain("mps")
+
+
+class TestExportPreflightBranch:
```
ChengHauYang, line 210: Following my other comment: with _NON_ACCELERATOR_FAMILIES, meta goes back to being just like cpu -- no toolchain, no crash. So this test flips: instead of expecting ValueError, patch check_toolchain (and stub the Inductor call) and assert it's never called for meta.

doc/content/references/aoti_packages.md
```diff
@@ -18,7 +18,7 @@ The CLI synopsis:
 ```
 neml2-compile <input.i> --model <name>
                         [--output-dir <dir>]
-                        [--device cpu|cuda [cpu|cuda ...]] [--dtype float64|float32]
+                        [--device cpu|cuda|xpu|hip|mps [...]] [--dtype float64|float32]
```
ChengHauYang, line 21: The help shows [--device cpu|cuda|xpu|hip|mps [...]] [--dtype float64|float32] like every combo works -- but mps doesn't support float64.

How about an early error out: right after parsing --device/--dtype, check family/dtype and say something like "mps only supports float32; pass --dtype float32".

ChengHauYang (commented):
Happy to see that neml2 can support more devices. Good work!

neml2/_accelerator.py
reverendbedford: This is a good approach.

tests/unit/test_accelerator.py
reverendbedford: Same, this is a better approach than what we have.

doc/content/references/aoti_packages.md
reverendbedford: I'll add device/dtype checking.  I'll error if the user *only* asks for an invalid combination and give a warning if they ask for a combination of valid + invalid.

neml2/_accelerator.py
reverendbedford: Done in de7102863 -- added `_NON_ACCELERATOR_FAMILIES = ("cpu", "meta")` in `neml2/_accelerator.py` and split `target_accelerator_family` into the three cases you outlined (skip cpu/meta, return known accelerators, raise `ValueError` for anything else so an unroutable family fails at the front instead of confusingly inside `check_toolchain`).

tests/unit/test_accelerator.py
reverendbedford: Done in de7102863. `TestExportPreflightBranch` now has two tests using `monkeypatch`: (a) accelerator branch fires when `target_accelerator_family` returns a real family (patched at the source so the deferred `from` import inside `compile_model` picks it up), and `check_toolchain` records the call before a sentinel exception short-circuits the real Inductor compile; (b) branch is skipped for a real meta tensor, proven by `check_toolchain` never being called and control flowing to the stubbed Inductor call.

doc/content/references/aoti_packages.md
reverendbedford: Done in de7102863 (Mark asked for this in-thread too). New `is_compatible(family, dtype)` in `neml2/_accelerator.py` -- allow-list keyed by family, only `mps + float64` is invalid today (families absent from the table stay permissive). `neml2-compile` filters the `--device` list against `--dtype`: warns "ignoring &lt;dev&gt; + &lt;dtype&gt;" on every dropped combination, and `parser.error(...)` when nothing valid remains. `neml2-run` (single device) errors straight out with an actionable message. Tests added for both CLI paths and the underlying predicate.

ChengHauYang (approved):
This PR looks good to me. 😃

